PR #1, opened 2016-06-13
Add simple test for utils

jchodera:


jchodera:
It looks like I'll need to move the `openeye` imports into the functions they are used to get this to work. The code should work without this testing framework, though!

PR #2, opened 2016-06-16
Added another example, being a AlkEthOH-based one

cbayly13:


davidlmobley:
@cbayly13 notes that the code change `oechem.OEWriteConstMolecule(ofs, mol)` is particular important and we should be sure that we use OEWriteConstMolecule everywhere. Otherwise, if the write is not the last operation, multiple writes of the same molecule could result in modifications to the molecules.

I will need to check everything in `openmoltools` to ensure that we do this.

davidlmobley:
@jchodera - you can now review this.

jchodera:
Looks good to me. Merge when ready!

PR #5, opened 2016-06-18
Fix bug where effective temperature was not passed to AtomTypeSampler

davidlmobley:
Testing at temperatures other than T=0.1 revealed that the effective temperature option did nothing. This fixes the issue. 

( @cbayly13 @bannanc )


davidlmobley:
@jchodera - happy with a merge? Very minor change.

jchodera:
Great catch! Thanks!

PR #6, opened 2016-06-20
two changes: Caitlin's csv trajectory writer and my "improved" decorators for AlkEthOH

cbayly13:


jchodera:
Did you mean to delete the encrypted openeye license? Travis testing will
not work without it!

davidlmobley:
That should get re-added, @cbayly13 and @bannanc .

jchodera:
A number of files also had their file modes strangely changed. Are you using some weird editors or filesystems or something?

davidlmobley:
He's using vi under OS X... I'll bring this up and see if I can figure anything out. Odd.

jchodera:
It probably doesn't impact anything, but it might cause unexpected behavior in the future.

Suggest [atom](https://atom.io/)

bannanc:
no we definitely did not! I have it back on my fork, but this pull request is from @cbayly13 's fork so I created a separate pull request from mine. It's identical to his right now (besides the missing license)

jchodera:
No need to wait for me to review.

bannanc:
@cbayly13 most recent commit now has the oe_license_encrypted file. @davidlmobley it should be ready to merge

davidlmobley:
@bannanc - checking.

davidlmobley:
@bannanc @cbayly13 - seems to be missing the temperature fixes from #5 . Investigating why this is not showing up in the change log.

davidlmobley:
@bannanc @cbayly13 - you want to make sure that the branch/fork you are PRing from is not behind master before submitting a PR. In this case on GitHub you'll see at the top of the page here: https://github.com/cbayly13/smarty
it says:

"This branch is 13 commits ahead, 4 commits behind open-forcefield-group:master."

So you're missing the last four commits to master, which were fixes to the temperature bug.

davidlmobley:
OK, they got it this time via the merge above. Now it's ready to be merged.

jchodera:
I just realized that this PR broke the `test_sampler.py` test, since you added a `trajFile` argument without updating the corresponding test.

I'll fix that in #24, but I suggest making things like `trajFile` optional arguments in a new PR (e.g. `trajFile=None`), and having the trajectory file only be written if a file object is specified here.

It's best to avoid introducing breaking the API by adding new arguments that are _not_ optional if it can be avoided.

davidlmobley:
This will also be avoided in the future now that we're switching back to
branches here rather than forks. We didn't realize the tests weren't
getting run until after this. ( @bannanc )

On Mon, Jun 27, 2016 at 9:48 PM, John Chodera notifications@github.com
wrote:

> I just realized that this PR broke the test_sampler.py test, since you
> added a trajFile argument without updating the corresponding test.
> 
> I'll fix that in #24
> https://github.com/open-forcefield-group/smarty/pull/24, but I suggest
> making things like trajFile optional arguments in a new PR (e.g.
> trajFile=None), and having the trajectory file only be written if a file
> object is specified here.
> 
> It's best to avoid introducing breaking the API by adding new arguments
> that are _not_ optional if it can be avoided.
> 
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub
> https://github.com/open-forcefield-group/smarty/pull/6#issuecomment-228947391,
> or mute the thread
> https://github.com/notifications/unsubscribe/AGzUYfXWiX4bWSoRkQt00AAoBQF3PxNVks5qQKemgaJpZM4I6BoU
> .

## 

David Mobley
dmobley@gmail.com
949-385-2436

jchodera:
I'll fix this in #24

bannanc:
Thanks John,
I'll be more careful with the tests in the future. Hopefully it won't be a problem now that we're switching to branches.

jchodera:
> Hopefully it won't be a problem now that we're switching to branches.

Awesome, I think that will solve this issue!

Issue #13, opened 2016-06-21
Allow binary rather than unary decorators for smarts sampling


davidlmobley:
@jchodera , this one will require your attention.

We believe that to allow better exploration we need to allow binary rather than unary construction of SMARTS to sample. I haven't dug in too much to the details of the code on this aspect, but what @cbayly13  is telling us on this end is that he needs to be able to combine "pick a bond order" with "pick an element type or functional group" more easily - i.e. rather than taking a decorator that does both at once, being able to combine two decorators. In other words, instead of a unary decorator, he wants binary decorators (take this action such as a bond, apply it to that chemical group). 

So I think he wants to be able to revise this:

```
$(*~[#1])      hydrogen-adjacent
$(*~[#6])      carbon-adjacent
$(*~[#7])      nitrogen-adjacent
$(*~[#8])      oxygen-adjacent
$(*~[#9])      fluorine-adjacent
$(*~[#15])     phosphorous-adjacent
$(*~[#16])     sulfur-adjacent
$(*~[#17])     chlorine-adjacent
$(*~[#35])     bromine-adjacent
$(*~[#53])     iodine-adjacent
```

by replacing it with "pick a bond type" as one operation and "pick a functional group adjacent to it" as a second.

@cbayly13 , have I summarized properly? 

Ultimately, he will also want to optimize how these selections are made under the hood, i.e. trying a single bond basically works with everything, but double or triple bonds only make sense for particular elements/functional groups, etc. This may be an API issue.

jchodera:
We could add a "binary decorators" category to our current "unary decorators" category, but this will significantly increase both the complexity and the rejection rate.

You can mimic this capability right now by manually expanding the list with sensible combinations. For example, you wouldn't want:

```
$(*-[#1])      hydrogen-adjacent-bondorder1
$(*=[#1])      hydrogen-adjacent-bondorder2
$(*#[#1])      hydrogen-adjacent-bondorder3
```

because most of those don't make sense, but you can try

```
$(*-[#1])      hydrogen-adjacent
$(*~[#6])      carbon-adjacent
$(*-[#6])      carbon-adjacent-bondorder1
$(*=[#6])      carbon-adjacent-bondorder2
$(*#[#6])      carbon-adjacent-bondorder3
$(*-[#7])      nitrogen-adjacent-bondorder1
$(*=[#7])      nitrogen-adjacent-bondorder2
$(*#[#7])      nitrogen-adjacent-bondorder3
[...]

Issue #12, opened 2016-06-21
Troubleshoot probable descriptor problem


davidlmobley:
We likely have a problem in what descriptors we are sampling, as in the AlkEthOH test, we are **never** able to recover the `H1`, `H2`, and `H3` sub-types of `HC` even in the case where we explore the `HC` branch first, it seems. This suggests there is a problem with whether/how suitable descriptors are generated in addition to the tree exploration issues raised in #8 

@cbayly13 is investigating.

davidlmobley:
We are working on a fix for this which should be address when we resolve #13 . There is no need to retain this as a separate issue.

Issue #8, opened 2016-06-20
Discussion: Hierarchical approach as currently implemented poses problems for sub-types


davidlmobley:
The AlkEthOH test set reveals some "sampling problems" which we believe need to be resolved by some design modifications. 

Specifically, the `hydrogen` element contains multiple sub-types in AlkEthOH - most obviously, `HC` and `HO` which are hydrogens with attached carbons and hydrogens with attached oxygens. However, there are additional sub-types `H1`, `H2`, and `H3` that are sub-types of `HC` relating to what other type of atom is connected. The tree looks something like this:

```
         hydrogen
        /         \
      HC           HO
       |
      H1
       | 
      H2
       | 
      H3
```

Currently, as we understand it, the scoring approach dictates `smarty` can only discover the `H1`, `H2`, and `H3` types if it first discovers `HC` (and we haven't seen them discovered yet regardless). This is because if `smarty` matches `HO` first, then, since all remaining types are covered by `HC`, discovering `HC` leads to no improvements to the score AND, **worse**, would lead to no atoms being typed by the generic `hydrogen` (so this automatically gets rejected since each pattern is required to match at least some types).

It seems clear that in the limit of infinite sampling at finite temperature, one ought to be able to explore both branches and find that only the "discover HC first" branch yields to achieving the best possible score. However, we think that:
a) this is an inefficient way of exploring
b) we can easily make the "greedy" algorithm work better, and
c) this will likely result in other problems in more complicated trees - i.e. if both branches extended downwards several levels, we would only be able to explore one or the other but not both (since the generic type is required to match at least some atoms).

We believe that at least two changes should be made to resolve this; I'll lay them out briefly here and then can follow up by creating issues on the individual items once we agree on the concept:
1) **Base atom types should be allowed to match no atoms** and there should be no benefit to the score for them to match atoms; these base atom types should get special treatment.
- Indeed, Christopher would argue that base atom types OUGHT to end up matching no atoms HERE; these out to be basically a catch-all for anything we can't recognize or don't know anything about yet
- This will help because then developing new specialized types `HC` and `HO` in the example above will result in score increases

**Postpone discussion of this point following, need to clarify something with Bayly.**
[...]

PR #22, opened 2016-06-22
Resolves basetypes issue of #8 by allowing basetypes with zero occupancy

davidlmobley:
This resolves the hierarchy issue discussed in #8 by allowing creation of derivative types that make the base types (and **only** the base types) match zero atoms. As discussed in #8, the goal we believe ought to be to find specialized atom types that cover all cases in the set, leaving the base types empty. 

This does NOT yet fix the AlkEthOH example so that it recovers the `H1`, `H2`, and `H3` types because of deficiencies in how descriptors are created/sampled ( #12 ) which we think will be fixed in part by allowing binary operations such as in #13 .

This PR also brings in a very minor fix to a doc string.

### DIFF
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -94,6 +94,10 @@ class AtomTypeSampler(object):
         self.decorators = AtomTyper.read_typelist(decorators_filename)
         self.replacements = AtomTyper.read_typelist(replacements_filename)
 
+        # Store a copy of the basetypes, as these (and only these) are allowed 
+        # to end up with zero occupancy
+        self.basetypes = copy.deepcopy(self.atomtypes)
+
         # Store a deep copy of the molecules since they will be annotated
         self.molecules = copy.deepcopy(molecules)
 
@@ -317,8 +321,8 @@ class AtomTypeSampler(object):
                     valid_proposal = False
                     # Store this atomtype to speed up future rejections
                     self.atomtypes_with_no_matches.add(proposed_atomtype)
-                # Reject if parent type is now unused.
-                if (proposed_atom_typecounts[atomtype_typename] == 0):
+                # Reject if parent type is now unused, UNLESS it is a base type
+                if (proposed_atom_typecounts[atomtype_typename] == 0) and (atomtype_typename not in self.basetypes):
                     # Reject because new type is unused in dataset.
                     if self.verbose: print("Parent type '%s' (%s) now unused in dataset; rejecting." % (atomtype, atomtype_typename))
                     valid_proposal = False
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -82,10 +82,6 @@ class AtomTypeSampler(object):
         -----
         This is just a proof of concept.  No scoring of molecular properties is performed.
 
-        TODO
-        ----
-        * Maintain a list of types that do not type any molecules so that we can avoid proposing these again.
-
         """
 
         self.verbose = verbose
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -97,6 +97,8 @@ class AtomTypeSampler(object):
         # Store a copy of the basetypes, as these (and only these) are allowed 
         # to end up with zero occupancy
         self.basetypes = copy.deepcopy(self.atomtypes)
+        # Store smarts for basetypes
+        self.basetypes_smarts = [ smarts for (smarts, name) in self.basetypes ]
 
         # Store a deep copy of the molecules since they will be annotated
         self.molecules = copy.deepcopy(molecules)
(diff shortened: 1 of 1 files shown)


davidlmobley:
@bannanc , code review/testing would be appreciated. Works for me, but I'd like you to think through it too.

jchodera:
Are you sure you only want to allow empty basetypes (types that were present initially), rather than empty _parents_?

jchodera:
Also, I'm a bit surprised that no tests were run!  Is it because we aren't running tests if the PR is from a fork (due to the OpenEye license requirements)?

```
$ cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=2 --with-doctest --with-timer -a "\!slow" && cd ..
Ran 0 tests in 0.001s
```

jchodera:
Perhaps we want to move to using [feature branches of the main repo](https://bocoup.com/weblog/git-workflow-walkthrough-feature-branches) to ensure tests are run in the future?

There is a way to [do this safely](https://help.github.com/articles/about-protected-branches/) by preventing pushes to the master.

davidlmobley:
@jchodera :

> Are you sure you only want to allow empty basetypes (types that were present initially), rather than empty parents?

Yes, we are sure that this is what we want to try for now. We are worried about the potential additional complexity that might run away if we allowed empty parents throughout. Possibly the data might drive us to that eventually, but we are not ready to go there yet, and it worries us. We're trying to be conservative with changes at least until we get to "success" on AlkEthOH. 

Where I'd like to end up is with automated tests on AlkEthOH and some similar small-scale tests so that we ensure that we can reach a score of 100% when we ought to be able to so that we can easily detect changes to the code which would result in us no longer being able to do so. 

> Also, I'm a bit surprised that no tests were run! Is it because we aren't running tests if the PR is from a fork (due to the OpenEye license requirements)?

I'm confused about this too. I thought PRs from forks in `openmoltools` were able to run OE tests? If not then we definitely should move to an alternate model, as we need useful tests here!

davidlmobley:
Also - Chris's argument is that empty **base types** is the goal, because if something is left as "generic carbon" it's trying to tell us that we don't know anything about it and we need to. However, the same argument would NOT apply to empty parents that are further down the tree. So I think the data would have to compel us to go there before we want to try that.

jchodera:
> I'm confused about this too. I thought PRs from forks in openmoltools were able to run OE tests? If not then we definitely should move to an alternate model, as we need useful tests here!

PRs from _branches_ within the `choderalab/openmoltools` can run tests. PRs from _forks_ cannot.

jchodera:
> Also - Chris's argument is that empty base types is the goal, because if something is left as "generic carbon" it's trying to tell us that we don't know anything about it and we need to. However, the same argument would NOT apply to empty parents that are further down the tree. So I think the data would have to compel us to go there before we want to try that.

We may need to separate basetypes from initial types in the future, then.  Right now, there's just a single input file to specify basetypes, which are the same as the initial types.

davidlmobley:
> PRs from branches within the choderalab/openmoltools can run tests. PRs from forks cannot.

Ah, right. Now I remember why I was confused - it's because `openmoltools` has useful tests which do not require the OpenEye tools, whereas we don't (and won't, I think) here. This means we probably need to switch to the branch model here. 

> We may need to separate basetypes from initial types in the future, then. Right now, there's just a single input file to specify basetypes, which are the same as the initial types.

Yes, agreed. I'll make an issue on that.

bannanc:
> > Also - Chris's argument is that empty base types is the goal, because if something is left as "generic carbon" it's trying to tell us that we don't know anything about it and we need to. However, the same argument would NOT apply to empty parents that are further down the tree. So I think the data would have to compel us to go there before we want to try that.
> 
> We may need to separate basetypes from initial types in the future, then. Right now, there's just a single input file to specify basetypes, which are the same as the initial types.

Following this logic, we will probably need to stop base types from being destroyed so we know when they're empty. 

@davidlmobley I pulled your changes. I'm running them and it seems to be working the way you intended. It mostly finds the same atom types that it did before, but with a more detailed smart string for it. All of the base types were destroyed. 

For example this was a final set of SMARTS strings 

```
INDEX        ATOMS  MOLECULES                                                          TYPE NAME                           SMARTS REF TYPE        FRACTION OF REF TY      PED MOLECULES MATCHED
     1 :        396         42 |                                       c_hydrogen total-h-count-0                          [#1&H0]       HC              244 /                    244 (100.000%)
     2 :         68         42 |                                       c_hydrogen oxygen-adjacent                   [#1&$(*~[#8])]       HO               68 /                     68 (100.000%)
     3 :        232         42 |                                       c_carbon hydrogen-adjacent                   [#6&$(*~[#1])]       CT              232 /                    232 (100.000%)
     4 :         39         30 |                                           c_oxygen connections-1                          [#8&X2]       OS               39 /                     39 (100.000%)
     5 :         68         42 |                           c_oxygen connections-1 total-h-count-1                       [#8&X2&H1]       OH               68 /                     68 (100.000%)
 TOTAL :        803         42 |                                                                                                         651 /      803 match (81.071       %)
```

davidlmobley:
Good point, @bannanc . I've put this on my list. Note it probably should be amended to say that we should stop basetypes which _initially_ match something from being destroyed, since we're happy to destroy the generics which will never match anything (because of the set composition) such as bromine.

bannanc:
> which initially match something from being destroyed, since we're happy to destroy the generics which will never match anything (because of the set composition) such as bromine.

yes definitely what I meant

Probably we need to remove base types that don't match anything to make a list of "initial types" that match atoms. Then allow initial types to have a zero score, but not be removed. 

This has brought up many good points, but I think the current changes can be merged in for now.

PR #24, opened 2016-06-27
SMIRFF ForceField implementation

jchodera:
This begins the SMIRFF ForceField implementation.

For discussion of format, see: https://github.com/open-forcefield-group/open-forcefield-data/issues/9

Currently, only SMIRFF bond types are implemented, but the implementation for other force terms is very simple and should proceed quickly. I want to do more testing to make sure the current machinery is working for bonds, however.
- [x] bond types
- [x] angle types
- [x] torsion types
- [x] Lennard-Jones types
- [x] charge assignment from `OEMol`s
- [x] tests on SMIRFF ffxml parsing
- [x] tests on small molecules
- [x] tests on mixtures
- [x] automatic unit handling


jchodera:
Note that this implementation doesn't yet handle constraints---all bonds are added as harmonic bonds for now.

I propose we add a general constraint scheme later, perhaps with an SMIRKS-styled syntax:

``` XML
<Constraints>
  <!-- bonds involving hydrogen -->
  <Constraint smirks="[#1:1]-[*:2]"/> <!-- add constraint between atoms 1 and 2 using equilibrium bond distance -->
  <!-- angles involving hydrogen -->
  <Constraint smirks="[#1:1]-[*:2]-[#1:3]"/> <!-- add constraint between atoms 1 and 3 using auto-calculated distance from equilibrium bond and angles -->
</Constraints>
```

jchodera:
I've tried to make adding a new force generator require only a minimal amount of work. For harmonic bond forces, this is:

``` python
class HarmonicBondGenerator(object):
    """A HarmonicBondGenerator constructs a HarmonicBondForce."""

    class SMARTYBondType(object):
        """A SMARTY bond type."""
        def __init__(self, node):
            self.smirks = _validateSMIRKS(node.attrib['smirks'], node=node)
            self.length = _convertParameterToNumber(node.attrib['length'])
            self.k = _convertParameterToNumber(node.attrib['k'])

    def __init__(self, forcefield):
        self.ff = forcefield
        self._bondtypes = list()

    def registerBond(self, node):
        """Register a SMARTY bondtype definition."""
        bond = HarmonicBondGenerator.SMARTYBondType(node)
        self._bondtypes.append(bond)

    @staticmethod
    def parseElement(element, ff):
        # Find existing force generator or create new one.
        existing = [f for f in ff._forces if isinstance(f, HarmonicBondGenerator)]
        if len(existing) == 0:
            generator = HarmonicBondGenerator(ff)
            ff.registerGenerator(generator)
        else:
            generator = existing[0]

        # Register all SMARTY bond definitions.
        for bond in element.findall('Bond'):
            generator.registerBond(bond)

    def createForce(self, system, topology, **kwargs):
        # Find existing force or create new one.
        existing = [system.getForce(i) for i in range(system.getNumForces())]
        existing = [f for f in existing if type(f) == openmm.HarmonicBondForce]
        if len(existing) == 0:
            force = openmm.HarmonicBondForce()
            system.addForce(force)
        else:
            force = existing[0]

        # Iterate over all defined bond types, allowing later matches to override earlier ones.
        bonds = ValenceDict()
        for bond in self._bondtypes:
            for atom_indices in topology.getSMIRKSMatches(bond.smirks):
                bonds[atom_indices] = bond

        # Add all bonds to the system.
        for (atom_indices, bond) in bonds.items():
            force.addBond(atom_indices[0], atom_indices[1], bond.length, bond.k)

parsers["HarmonicBondForce"] = HarmonicBondGenerator.parseElement
```

I'll next write these for `HarmonicAngleForce`, `PeriodicTorsionForce`, and `NonbondedForce`.

jchodera:
I'll also note that the current `createSystem` function signature is

``` python
system = forcefield.createSystem(topology, mols)
```

where `mols` is a list of `openeye.oechem.OEGraphMol`s that specify the unique "reference molecules" present in the system. These reference molecules will also carry the partial charges used.

I can change this to SDF if that is preferred, but is there additional information we need in the tags that must be preserved in this case?

jchodera:
Matching of molecules identified in the `Topology` object to the molecules is currently done using [`networkx`](https://networkx.github.io/) based on graph isomorphism, using only elements and bond connectivity. I'm not sure if this has the potential to match the molecules incorrectly, but if it does, we would have to figure out what other properties to match on.

jchodera:
I can finish this PR up tonight, but it would be great to get comments from @davidlmobley and @pgrinaway (who has been asking for this for some time).

The biggest issues I can see right now are:
- The charges will have to be taken from the provided `OEMol`s, which is fine for now, but will eventually have to be replaced by a charging scheme.
- There is no understanding of polymeric entities. If you have a protein that consists of many amino acids, the valence and Lennard-Jones terms will be typed, but you still need to provide an `OEMol` for the whole protein from which the charges are taken. I'm also not sure how well the [NetworkX graph isomorphism algorithm](https://networkx.github.io/documentation/networkx-1.10/reference/algorithms.isomorphism.vf2.html) scales for very large molecules (subgraphs) like proteins.

Solving these issues will probably require we (1) modify the `Topology` format so that it is easy to construct an `OEMol` from the information in there (e.g. by adding bond orders), and (2) add a fragmentation algorithm that is aware of conjugation to fragment large molecules into smaller fragments in a way that can both allow facile charging and still be searched with SMIRKS. But these are longer-term issues.

davidlmobley:
@jchodera :

> ...where mols is a list of openeye.oechem.OEGraphMols that specify the unique "reference molecules" present in the system. These reference molecules will also carry the partial charges used.
> I can change this to SDF if that is preferred, but is there additional information we need in the tags that must be preserved in this case?

I'm not totally sure what you're asking about tags, since if you change it to taking in SDF files, you'd be pulling in whatever tags are in the SDF files so I don't see a significant difference (in terms of the result) than what you'd get pulling in info from OEMols. But, I'm assuming you're asking what would be in the tags. It would be something like: 
- `partial_bond_orders`
- `partial_charges`
- `atom_types` (we probably want to track these just for, er, "historical" reasons at present, though obviously they're going to be unimportant to the force field)
  However, Christopher Bayly notes that these names should be revised so they are very certain to never clash with any tags anyone else would create, to avoid unintended behaviors/clashes. He suggests adding the origin to the tags to achieve this, so maybe something more like:
- `OpenEyeAM1BCC_partial_charges`
- `OpenEyeAM1_partial_bond_orders`
- `antechamber_gaff_atom_types` and possibly `antechamber_sybyl_atom_types`
  (the last one I'm pretty ambivalent on since we don't anticipate significant functional importance to the atom types, but we may need these - especially sybyl atom types - down the road to get back into file formats which require atom types, such as .mol2)

In the interests of getting going quickly, I'm really OK with just having everything come in as OEMols for now, even though as we discussed on the phone last week there's some preference for taking SDFs. 

In terms of constraints - presumably we can get by without constraints for everything in AlkEthOH for now. It does have water, but that could be treated as flexible water (which would be more in keeping with the spirit of what we're doing right now anyway). 

Otherwise, I think this looks good to me (but _do you want me to try and do code review_?). 

Thinking of testing - I was thinking a sensible test for me to run once you think you have this working would be to do energy evaluations in OpenMM for the provided AlkEthOH conformations via parameters coming in from `.prmtop` and `.crd` files versus from this implementation of SMIRFF. This should check two things at once:
- Has Bayly input his SMIRFF file properly
- Are the parameters parsed properly to result in the same energies

Thoughts?

jchodera:
Thanks! Would be great if you could also look through the code to see if it makes sense or has appropriate documentation, since blocks of it will be replicated and modified next.

Do you have any input files I could use for more thorough testing? I am only testing against single molecules in vacuum right now.

davidlmobley:
@jchodera - I'll review the code, though it might not be til late. 

Input files - anything in particular you want to look at? Solvation Toolkit should be able to set up any mixture you want pretty easily via an API we discussed previously (https://github.com/mobleylab/solvationtoolkit) and I _almost_ have it conda installable (https://github.com/omnia-md/conda-recipes/pull/540 - I just have to bypass a couple software clashes I think). 

I'll generate a couple of simple mixtures of alkanes or cyclohexane or something for you and attach them momentarily.

jchodera:
> I'll generate a couple of simple mixtures of alkanes or cyclohexane or something for you and attach them momentarily.

Great!  A small number of "things that should work" would be great.

davidlmobley:
OK, @jchodera - attached please find mixtures of (cyclohexane, ethanol) and (propane, methanol, butanol). These are (40%,60%) and (20%, 30%, 50%) respectively. 

Within the data directory, you'll find subdirectories for `monomers`, which has the monomers in `.mol2` and `.sdf` formats; then there is a `packmol_boxes` directory with PDBs of the solvated boxes. These are out of packmol, and are cubic. (Note I also have AMBER .prmtop and .crd files there if it's easier for you to load those than PDBs; those are in the `amber` directory).

I expect you can just load:
- `data/monomers/ethanol.sdf` and `cyclohexane.sdf` and `data/packmol_boxes/cyclohexane_ethanol_0.4_0.6.pdb`
- `data/monomers/methane.sdf`, `propane.sdf`, and `butanol.sdf` and `data/packmol_boxes/propane_methane_butanol_0.2_0.3_0.5.pdb`

As noted if it's easier for you to get a `Topology` from AMBER files those are there as well. 

If you'd like anything else generated, I can easily do so. For example, I could blast through AlkEthOH and generate mixtures of every molecule paired with another molecule, or something along those lines. LMK. 

[data.tar.gz](https://github.com/open-forcefield-group/smarty/files/335654/data.tar.gz)

jchodera:
Thanks! This is a great start. Let me make sure this works first.

jchodera:
@davidlmobley: I've implemented tests based on the nice dataset you provided, but this has uncovered a problem in SMIRKS typing the OH bond in ethanol.

I'm using the `monomers/ethanol.mol2` file as a source for the ethanol molecule template, and while I can type C-H with `[#6:1]~[#1:2]` just fine, I can't match O-H with `[#8:1]~[#1:2]`.

See the [test code here](https://github.com/open-forcefield-group/smarty/blob/forcefield/smarty/tests/test_forcefield.py#L54-L92), which succeeds with `[#6:1]~[#1:2]`,

```
[LSKI1497:open-forcefield-group/smarty/smarty.open-forcefield-group] choderaj% python smarty/tests/test_forcefield.py
1 reference molecules loaded
('C1', 'H1')
('C1', 'H2')
('C1', 'H3')
('C2', 'H4')
('C2', 'H5')
```

but fails with `[#8:1]~[#6:2]`

```
[LSKI1497:open-forcefield-group/smarty/smarty.open-forcefield-group] choderaj% python smarty/tests/test_forcefield.py
1 reference molecules loaded
  File "smarty/tests/test_forcefield.py", line 96, in <module>
    test_smirks()
  File "smarty/tests/test_forcefield.py", line 92, in test_smirks
    raise Exception("Pattern '%s' did not match any molecules in set" % (smirks))
Exception: Pattern '[#8:1]~[#1:2]' did not match any molecules in set
```

Perhaps the SMIRKS syntax is incorrect here?

jchodera:
Another clue here: If I change to using the SDF files, it _does_ seem to perceive this bond, but (of course) the atom names are empty:

```
[LSKI1497:open-forcefield-group/smarty/smarty.open-forcefield-group] choderaj% python smarty/tests/test_forcefield.py
1 reference molecules loaded
('', '')
```

davidlmobley:
@jchodera : 

> I'm using the monomers/ethanol.mol2 file as a source for the ethanol molecule template, and while I can type C-H with [#6:1]~[#1:2] just fine, I can't match O-H with [#8:1]~[#1:2].

I expect this is because the `.mol2` files use GAFF atom types (I should have emphasized the importance of this when I was passing them on to you) so it's probably not perceiving the elements properly. You MAY be able to fix this by using the forcefield-flavor reader as we had to use the other week when working on Smarty. Alternatively, you can use `antechamber` to get back SYBYL types. (To test, you could write your OEMols back to `.mol2` files and see how the atoms are typed. I'm guessing the oxygens might be listed as "Os" or something.)

jchodera:
Aha! Thanks!

davidlmobley:
I also read through the code and it generally looked good to me though there were some things which are hard to follow, not knowing the inner workings of OpenMM. BUT, I did also see a couple potential issues:
- Line 96 of `forcefield.py`: `resname = molecule.GetTitle()`; is this going to be important for anything? If so we'll need to check that there is actually a title... What if it is blank? Won't line 97 fail? Or what if it is non-unique? (i.e. multiple molecules have the same residue name?) We have some code in `openmoltools` which ensures no mixtures are set up involving monomers which have the same residue names when using `packmol` as this causes failures elsewhere, but we may need something similar here. Inevitably someone is going to feed in molecules with no titles and/or molecules with non-unique residue names, I would imagine. 
- Aren't we calling the atom type sampler Smarty and the forcefield Smirff? So this code might need a nomenclature change: """Register a SMARTY bondtype definition."""
  `bond =HarmonicBondGenerator.SMARTYBondType(node)`

jchodera:
Thanks for the comments!
- `generateTopologyFromOEMol()` function is actually a duplicate from `openmoltools`, and is just a utility function. The `resname` isn't used anywhere, and this shouldn't fail if the molecule title is blank.
- Good point about SMIRFF. I'll make that change.

jchodera:
OK, all tests pass locally now!

Extending the forcefield parsing to handle the rest of the forces will have to wait until tomorrow.

@pgrinaway: Would still love input/feedback if you had any!

pgrinaway:
I think this is looking really great! It will be nice to have networkX as a part of this. The one minor comment that I will say is that it's best to stick with python style (`method_name` not `methodName`). But aside from that very minor comment, I think this is great!

jchodera:
Thanks, @pgrinaway! Will make a note to adopt PEP8 syntax, but for now, there are advantages to keeping the same method naming scheme as the original `simtk.openmm.app.ForceField` and trying to make it as similar as possible.

jchodera:
I've added angles, torsions, and Lennard-Jones parameters. This is just missing charges, which I'll add tomorrow.

I _may_ have some logic incorrect with torsions. Currently, all torsions are assumed to be symmetric so that matches for atom indices `(i,j,k,l)` and `(l,k,j,i)` are the same thing, which may not be true for improper torsions. Perhaps someone could comment on whether that logic?

davidlmobley:
@jchodera :

> I may have some logic incorrect with torsions. Currently, all torsions are assumed to be symmetric so that matches for atom indices (i,j,k,l) and (l,k,j,i) are the same thing, which may not be true for improper torsions. Perhaps someone could comment on whether that logic?

Weren't you the one who told me that impropers aren't symmetric? I thought you said permuting the atom order can be very important? 

@swails I'm sure could straighten us out... Jason, in AMBER-style force fields, are impropers `(i,j,k,l)` and `(l,k,j,i)` the same typically (symmetric) or would the order result in important differences?

Regardless, I would guess that this would not be an issue immediately as there are probably no impropers in AlkEthOH. @bannanc - maybe you can check the info you have on parm@frosst and see if you can tell anything on this? 

One thing I think we'll be in a position to do, @jchodera , is to soon cross-check the energies produced by this with the direct `tleap` implementation of parm@frosst in the AlkEthOH set. I can probably add part of that (not the whole thing) as a test once we do so.

swails:
I tried to get caught up on the discussion happening here, but I'm afraid I have only a partial picture.

As for the impropers in Amber, there are two different "symmetries" in atom indices of note here.  The first is the one that tleap uses when assigning impropers.  The second is the transformations that result in identical energies within the prmtop.  The following is how I _believe_ that tleap assigns impropers:
1. It definitely reorders all of the improper atom types (upon reading) to be in alphabetical order (by type, obviously), with wild-cards coming first (see `zParmSetOrderImproperAtoms` in `parmSet.c`).
2. At potential improper centers (not sure how this is defined), tleap checks various permutations of the atom types involved in the potential improper and sees if there is an exact or wild-card match in the database.  If it finds one, it puts the central atom in the 3rd position and orders the remaining 3 atoms (in positions 1, 2, and 4) according to atom type name (and, I believe, possibly residue number).  So it's possible you'd see some permutations of i, j, k, l where i, j, and l can be in any order if the atom type name and residue number are identical (indeed, I've seen prmtop files where identical residues can have different orderings since tleap does not use a stable sort for that).

Once it's in the prmtop, the energy expression is equivalent for i-j-k-l and l-k-j-i.  tleap swaps the order of improper torsions IFF the first atom in the system (which has an index of 0) is either in position k or l, since impropers assign the 3rd and 4th indices to negative numbers to indicate exclusion from the 1-4 list and that it is an improper.  Since -0 == 0, the first atom has to come in one of the first two positions.

I'm not sure if this answered your question or not... Assigning impropers is confusing for all packages :(

jchodera:
@swails: Thanks so much for the lucid clarification! 

The code is now done, but there are two issues we should sort out with @cbayly13:

### Ambiguity in impropers

There is a real issue that @swails brings up regarding the ill-determinedness of impropers.

Consider an improper that keeps the nitrogen in a peptide bond planar.  We could specify a specific enough SMIRKS string that there is no ambiguity in the improper, such as `[C:1](=O)-[N:3](H:4)-[C:2]`, but there is also a way to specify the SMIRKS string so that the improper is ambiguous: (e.g. `[C:1]-[N:3](H:4)-[C:2]`, where the carbons connected to the nitrogen can be matched in either way). In this case, the SMIRKS would match both ways, and we would accidentally apply the torsion twice. So I should probably add error checking to make sure the same _set_ of atoms isn't accidentally matched twice, throwing an exception if this happens---presuming we want to avoid ambiguity in impropers.

### Units for SMIRFF XML format

@cbayly13 didn't specify the units in his [SMIRFF XML example](https://github.com/open-forcefield-group/smarty/blob/b69f142a58a47bdcfca4efd3c13d49a6640ac9c8/smarty/data/forcefield/Frosst_AlkEtOH.ffxml), so I've had to guess these corresponded to the standard Amber choices. I'm not sure I guessed correctly---in fact, I'm pretty sure I didn't. The energies are finite, but they are weird, and you guys can help fix that with further testing.

What do you think about proposing a way to attach units to the file to have them automatically converted from whatever you like?  We could extend the XML format to include units:

``` XML
<HarmonicBondForce length="angstroms" k="kilocalories_per_mole/angstroms">
   <!-- length is in angstroms; k is in kcal/mol/angstrom -->
   <Bond smirks="[#6X4:1]-[#6X4:2]" length="1.626" k="310.0"/> <!-- CT-CT from frcmod.Frosst_AlkEthOH -->
...
</HarmonicBondForce>
```

or we could even specify units in a per-attribute basis:

``` XML
<HarmonicBondForce>
   <!-- length is in angstroms; k is in kcal/mol/angstrom -->
   <Bond smirks="[#6X4:1]-[#6X4:2]" length="1.626*angstroms" k="310.0*kilocalories_per_mole/angstroms"/> <!-- CT-CT from frcmod.Frosst_AlkEthOH -->
...
</HarmonicBondForce>
```

davidlmobley:
@bannanc - can you cross-check these files with the parameters in the AlkEthOH test set here https://github.com/open-forcefield-group/open-forcefield-data/tree/master/Model-Systems/AlkEthOH_distrib (and perhaps see also the AMBER input files in this PR https://github.com/open-forcefield-group/open-forcefield-data/pull/17 and let us know the whether the units are the same/different from what is in the AMBER files/frcmod files?)

davidlmobley:
@jchodera : I like this style:

> ``` XML
> <HarmonicBondForce length="angstroms" k="kilocalories_per_mole/angstroms">
> ```

I DEFINITELY think these files should have units. If we get to design our own format, it ABSOLUTELY should have units. That's one of my pet peeves about existing simulation file formats - they don't carry units so you have to go digging through the docs to even find out what units things are in, which is so annoying. Should have thought of it before...

jchodera:
OK, let me add the units handling before we merge this.

bannanc:
@jchodera and @davidlmobley 

I'm comparing values in the Frosst_AlkEtOH.ffxml file with the Amber force field file for AlkEthOH set (frcmod.Frosst_AlkEtOH [here](https://github.com/open-forcefield-group/open-forcefield-data/blob/master/Model-Systems/AlkEthOH_distrib/frcmod.Frosst_AlkEthOH)) 

They appear to have the same values in the majority of cases so I would conclude everything is in AMBER units. 

However, for the bond length for CT-CT in the frcmod file is 1.526 and it's 1.626 in the xml file. 
Also, unless I'm missing something, in the xml file there are no sigma and epsilon values for carbons or oxygens.

davidlmobley:
@bannanc - @cbayly13 hand-typed the XML file so probably he's mis-typed `1.526` as `1.626` I would imagine. 

@jchodera - does the sigma and epsilon issue explain your weird energies? Do you have carbons and oxygens without sigmas and epsilons?

davidlmobley:
**REGARDING MISSING SIGMAS AND EPSILONS**: It looks like the `.ffxml` file here is only the "abbreviated" version drawn from John's example here https://github.com/open-forcefield-group/open-forcefield-data/issues/9#issue-154386766 and one of us, _NOT_ @jchodera , should go ahead and manually convert the rest of this file https://github.com/open-forcefield-group/open-forcefield-data/blob/master/Model-Systems/SMIRFF_prototype/smarff_mockup1.Frosst_AlkEthOH into the `.ffxml` format. (Note that there are also missing entries for torsions, etc - basically every where he had a "..." in his example implementation there is stuff missing (torsions and nonbonded only, it looks like).)

There is no reason for @jchodera to do this conversion, though, as we can easily do that on this end. 

@jchodera - does this explain your weird energies?

jchodera:
Thanks, everyone!

Let me just get the units thing in and we can merge this, then you can fix up the `.ffxml` file.

jchodera:
Types are now implemented.

I imagine there is still a lot we can do to provide more useful error feedback when there are typos or mistakes in the `ffxml` file, but we can add those after this PR is merged or as we find they are needed to help avoid common mistakes.

jchodera:
Types are specified using this format:

``` XML
<PeriodicTorsionForce phase_unit="degrees" k_unit="kilocalories_per_mole">
...
```

Note that not all types have a corresponding `type_unit` attribute. Those are assumed to be dimensionless, so it's important to remember that if you accidentally make a typo and use, for example, `phrase_unit="degrees"` instead, it will assume the `phase` is unitless or in natural OpenMM units (radians).  So we can probably harden this further by enforcing units be specified when we expect them to be.

This should be ready to merge when all tests pass.

davidlmobley:
One other thought for the record (for me to think about later) - we need to add something which will catch cases where we just don't have parameters defined for some of the atoms, i.e. we should never be able to run a system where we just don't have LJ parameters for some of the atoms. Presumably we should add some sort of exception handling to catch this...

jchodera:
I can certainly add that. I just took out some code that looked for bonds
in the Topology that didn't have HarmonicBondForce terms.

We should probably have an option to do thorough checking. But we can add
in a separate PR. If you can come up with what we should check for in an
issue, I can quickly add it in a separate PR.

davidlmobley:
I'll get back to that, @jchodera . So far this looks good, so I'd suggest you merge once tests pass and we (on my end) can work on testing more extensively (and fix the issue with the incomplete `ffxml` file) very soon.

Issue #27, opened 2016-06-28
Prevent base types which initially match something from being destroyed


davidlmobley:
As per https://github.com/open-forcefield-group/smarty/pull/22#issuecomment-228490056 we should prevent base types which initially match something from being destroyed, though they can become un-populated.

davidlmobley:
Closed via #29 .

Issue #26, opened 2016-06-28
Allow separate specification of base types and initial types 


davidlmobley:
Since we plan to not allow base types to be destroyed (except those which don't match anything), we need to allow _base types_ and _initial types_ to be specified separately, as per https://github.com/open-forcefield-group/smarty/pull/22#issuecomment-227919231

PR #29, opened 2016-06-30
Resolve issue 27 by retaining generic base types

davidlmobley:
This should resolve issue #27 by retaining base types which initially match something, even if they are unpopulated. The idea is that ultimately we will end up with generic elemental base types which match nothing but are still retained as catch-alls. 

I still separately have to deal with #26 to allow separate specification of base (generic) types and initial types, as these are now two different ideas. 

@bannanc , can you review?

### DIFF
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -102,6 +102,12 @@ class AtomTypeSampler(object):
         # Store smarts for basetypes
         self.basetypes_smarts = [ smarts for (smarts, name) in self.basetypes ]
 
+        # Store initially populated base types, as these will be retained even 
+        # if they have zero occupancy (whereas unpopulated base types
+        # need never be used ever and can be deleted- i.e. if we have no 
+        # phosphorous in the set we don't need a phosphorous base type)
+        self.used_basetypes = []
+
         # Store a deep copy of the molecules since they will be annotated
         self.molecules = copy.deepcopy(molecules)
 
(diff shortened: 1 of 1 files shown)


jchodera:
I don't think you want to equate the initial sampler state with basetypes, since this gives the initial sampler state some sort of privileged meaning. Part of our goal is to do experiments to see if starting from different sampler states can reach the same region, but that may be impossible if the rules of what can be accepted or rejected are different for different initial conditions.

I think we want to separate the initial sampler state definition from the basetypes definition, and only retain basetypes.

Even better would be to only retain atomtypes that have children, since this generalizes the concept of basetypes.

davidlmobley:
> I don't think you want to equate the initial sampler state with basetypes, since this gives the initial sampler state some sort of privileged meaning. Part of our goal is to do experiments to see if starting from different sampler states can reach the same region, but that may be impossible if the rules of what can be accepted or rejected are different for different initial conditions.

@jchodera : Yes, exactly. We already discussed this, and **that's issue #26 which I was going to do next**. :) I just ended up with only very little time to work on this yesterday so I couldn't do both at once as it required me spending time wrangling file parsing which was part of the code I hadn't worked through yet. 

The place where we want to end up is that we have "generic types" which we will always retain which will often be empty, deliberately, and we have separate "initial types" which are just the starting point.

davidlmobley:
@jchodera : 

> Even better would be to only retain atomtypes that have children, since this generalizes the concept of basetypes.

I'm quite confident this is NOT what we're after. For example, this would mean that (currently) in AlkEthOH we would have only one type of hydrogen, the generic type, since neither 'HO' nor 'HC' have children (yet - but they will once we get the SMARTS sampling working well enough). Also, our view (with Christopher) is that we want the generic/basetypes to normally go completely or nearly unused so that they can sort of serve as a flag for "we don't yet know much/anything about this chemistry". So we are actually trying to deliberately avoid generalizing the concept of basetypes and really treat the elemental basetypes in a special way.

jchodera:
OK, then let's at least split up the basetypes file and the initial types file into two separate entities, or else we will be preventing ourselves from being able to address one of the major experimental goals for this codebase (which is to examine the dependence of convergence on initial sampler state).

davidlmobley:
@jchodera - yup, handling that in #26 . :)

bannanc:
I just finished looking through the changes with this pull request, I think the code looks good. @davidlmobley

Issue #34, opened 2016-07-01
Implementing Hierarchy Tracking


bannanc:
We would like a way to track hierarchy (parents and children) so we can make tree diagrams and track how smarty is making changes. 

My plan currently is to create a dictionary which will track a list of children for each parent. 

@davidlmobley and I discussed creating objects for chemical environments. However, we are concerned that this would have to be completely redone when we move to sampling bonds, angles, and torsions.

jchodera:
Currently, we track a tuple of `(smarts, typename)`. What about just extending this to `(smarts, typename, parent)`?

jchodera:
From that information, you can completely reconstruct the (directed) graph of which types gave rise to which children (since we cannot eliminate parent types).

davidlmobley:
@jchodera : Is there a strong reason to prefer that approach over just providing lookup of parent `(smarts, typename)` via something like a `parents` dictionary? Extending the tuple would require much more extensive code changes as everything which handles a two-element tuple would need modification, whereas if we just have look-up when needed via a dictionary code changes will be isolated.

We figured we could use this dictionary approach for now, since down the line (when we move to bonds, angles, and torsions) we'll need more extensive rewrites anyway and that would be a good time to extend to a more general notion of chemical environment/parent tracking.

jchodera:
I don't think you save much in going the dict approach, since you need to modify the dict and pass it around in the sample places as the tuple, but go ahead and give it a try!

bannanc:
I'm looking through the code again, I've only interacted with a few parts of the sampler. If we're going to have to pass the dictionary around then it defeats the point of introducing it.

davidlmobley:
I think there's quite a bit of the code which interacts with types without modifying them or needing to know about parents/children, i.e. things printing type statistics, the part which actually does the type matching, and so on. But yeah, @bannanc can sort that out.

jchodera:
>  If we're going to have to pass the dictionary around then it defeats the point of introducing it.

That's what I was thinking.  But I'm not sure what is easiest.

davidlmobley:
Closed by #35

PR #35, opened 2016-07-01
adding parent dictionary and output of hierarchy tree

bannanc:
Adds "parent" dictionary to track parents and children 

Since there are many methods that use the atom list, I decided it would be easier to introduce a dictionary since it only needs to be used in the implementation and the sample_atomtypes method. 

My next plan is to print the tree (hierarchy) at the end of a run, if it's verbose, but I haven't quite sorted the easiest way to do that yet. 

Addresses issue #34

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,57 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# PyInstaller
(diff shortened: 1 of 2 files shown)


bannanc:
@davidlmobley 

Most recent commit prints 'tree' at the end of the run with output like this:

```
Atom type hierarchy:
[#1]
|   [#1&$(*~[#6])]
[#8]
|   [#8&H0]
[#6]
```

Later we can add something prettier, but it at least lets us see the hierarchy for now.

bannanc:
@davidlmobley I added a gitignore file since this was still open. 
I think this is ready to go unless @jchodera wants to look at it.

jchodera:
Looks good to me!

davidlmobley:
Looks good to me, @bannanc . Maybe add a low priority issue (enhancement) to allow visualization of the tree via that python module noted elsewhere; it's not important now but eventually when we get to parm@frosst on the ZINC set it will probably be helpful. 

This also resolves issue #36 .

PR #39, opened 2016-07-03
Implementing binary decorators for smarts sampling

camizanette:
I started to implement the idea that @cbayly13 gave to me about having binary decorators instead of unary decorators for smarty.
The first modification was changing the decorator file, where we removed the bond-atom specification and let only bond specifications.

```
$(*-z) simply bonded
$(*=z) doubly bonded
$(*#z) triply bonded
$(*:z) aromatic
$(*~z) any 
```

Then, we started to generate smarts strings randomly adding one, two or three decorators, as we discussed on #13. The modification was mostly inside the function sample_atomtypes.

The code is not done yet, and I need to clean up a lot of things, but we can have an idea about what is going on and what we can improve from that.

Please, let me know if you need more explanation about it.


jchodera:
@camizanette: This sounds like _exactly_ what we need. Thanks for putting this PR together, and succinctly summarizing the proposed solution in #13!

Some comments on the PR in progress:
- What do you think about retaining both the old behavior and new behavior? It might be simplest to add an option (e.g. a command-line option) to select a mode (perhaps `simple-decorators` and `combinatorial-decorators`), and to also refactor the child atomtype proposal into separate functions that handle the different behaviors. That way, we could easily experiment with adding other schemes in the future without changing the existing code.
- It looks like you accidentally ran `git add` on a number of directories and files that shouldn't be in the repo:
  - `smarty.egg-info/`
  - `build/`

You'll have to remove them, since it would cause trouble to have these committed to the repo since they are rebuilt each time you install.

Again, this is looking great---thanks for working on this!

bannanc:
@camizanette , @davidlmobley asked me to take a look at your pull request. I found one small error that is due to variable name differences. 

```
            # Update proposed parent dictionary
            proposed_parents[atomtype].append(proposed_atomtype)
            proposed_parents[proposed_atomtype] = []
```

The lines above aren't working anymore, I think it's because you've changed variable names. I added them while you were working on this to track parent hierarchies. 
atomtype was the atom that was being decorated, the variable doesn't seem to exist anymore. Are you calling that basetype now? 

Also, I think you'll want to use print(...) for your print statements. 

> What do you think about retaining both the old behavior and new behavior? It might be simplest to add an option (e.g. a command-line option) to select a mode (perhaps simple-decorators and combinatorial-decorators), and to also refactor the child atomtype proposal into separate functions that handle the different behaviors. That way, we could easily experiment with adding other schemes in the future without changing the existing code.

I like this idea, I think it will give us flexibility in experimenting later. What do you think?

davidlmobley:
@camizanette - after discussion with @cbayly13 and @bannanc , we're hoping you can go ahead and get the minor issues noted above (plus the optional command-line option) fixed so we can go ahead and merge this and begin using it more, ASAP. @bannanc is I think going to coordinate with you about who fixes the `proposed_parents` issue she mentioned.

We've got some more extensive changes that we want to make that I'll describe in a separate issue shortly, but we should get this merged **before** doing that.

bannanc:
@camizanette I fixed the parent issue, I'll make a pull request to your branch.

camizanette:
Thank you all for the comments. I updated my branch and I will submit the pull request tomorrow morning.

bannanc:
@camizanette I don't think this handling the parents/children correctly still 

For example, in a run I just did I got the following for children of hydrogen:

```
[#1]
         [#6$(*~[#1])+0]
         [#1H0]
```

The first child is a carbon not a hydrogen. However the second one is a hydrogen so I think I'm missing something about which variable you're using to make the children.

camizanette:
Thanks, @bannanc . Working on this!

camizanette:
@bannanc, can you check if it is correct now? Thanks!

davidlmobley:
If I understand right, this is good to merge now once @bannanc and/or I have a look? If so you can remove [WIP] from the title, @camizanette .

bannanc:
The parents look good to me now. Here's the tree from my last run:

```
[#1]
         [#1$(*~[#8])X1]
[#8]
         [#8$(*~[#6])H1]
[#6]
```

davidlmobley:
Thanks! So looks like I can merge... I'll go ahead and do so. Then we can shift our focus to #48 . It's nice to see that we're finding more complicated/decorated smarts patterns now.

It would probably be good if we understood why we can't find `H1` yet, but maybe we can get an idea of that when we talk this afternoon.

jchodera:
Awesome!

PR #47, opened 2016-07-05
fix parent dictionary

bannanc:
I fixed the parent dictionary to use your variable name. 

I also removed the build/ and smarty-egg.info/ directories. I don't know if that will help it pass the tests or not.


davidlmobley:
I think this got resolved in the open PR, @bannanc ? So we should close?

bannanc:
these changes are now included in pull request #39 So I'm closing this

Issue #56, opened 2016-07-09
Fix sporadic failure (key error) introduced in last PR


davidlmobley:
It looks like #55 introduced a sporadic Smarty failure, i.e. a key error is possible in some situations but the tests aren't long enough to always catch it - at least that's my take on the error messages (i.e. see https://travis-ci.org/open-forcefield-group/smarty/jobs/143498708 - failures in two out of three cases). 

@bannanc , can you take a look?

Thanks.

davidlmobley:
@bannanc - just wanted to make sure you saw this? 

One suggestion to aid troubleshooting is to make sure to fix the random seed somewhere to allow reproducing the problem.

bannanc:
@davidlmobley I just saw this I'll take a look when I get into the office this morning.

bannanc:
So the tests are still passing locally, but I'm getting a key error when I run with more iterations, I'm investigating now.

bannanc:
@davidlmobley  
I think I've found the problem. I was under the impression we were removing "unused" base types from the self.basetype list, but those are in a separate list self.used_basetypes. 

This means my code was looking for something like chlorine in the atomlist after it had been removed since it is still in basetypes. 

I'm doing some tests to double check, I'll get a pull request up soon.

PR #55, opened 2016-07-08
Added a method for focused atom searching

bannanc:
While this method might not get used beyond AlkEthOH, we believe it will help trouble shooting. WE should have new atom_sampler methods soon. 

It lead me to re-organize the parent dictionaries to include SMARTS and 'type-names'. 
There's also a bit of a reorganization for the run method (which also makes the use of the trajectory file a little more logical). 

This addresses issues #54 and #50

PR #57, opened 2016-07-11
Fixing bug with basetypes vs. used_basetypes

bannanc:
This addresses a bug discussed in issue #56  . 

Pull request #55 introduced a method to track which elemental atom type "trees" were complete. This used the atom type list `self.basetypes` when it should have been using `self.unused_basetypes` 

This caused a key error when atoms such as chlorine were removed from the atom list, but were still in `self.basetypes`

I also fixed a typo in decorators where a single bond was described as `simply-bonded` to `singly-bonded` 

I also noticed in this log file that there is a print statement with the parent dictionary, that I imagine was helpful for testing, but shouldn't need to be there now. It uses `str(self.parents)` which is hard to look at. We could print the parent tree at every step, with the established code if we want to see it. @camizanette and @davidlmobley do you have preferences on this last point? I can add it to this update.

### DIFF
--- a/examples/AlkEtOH2/atomtypes/decorators.smarts
+++ b/examples/AlkEtOH2/atomtypes/decorators.smarts
@@ -1,5 +1,5 @@
 % bonded to atoms
-$(*-z) simply-bonded
+$(*-z) singly-bonded
 $(*=z) doubly-bonded
 $(*#z) triply-bonded
 $(*:z) aromatic-bond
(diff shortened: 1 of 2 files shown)


davidlmobley:
@bannanc - printing the parent tree using the established code if we ask for verbose output sounds good to me.

bannanc:
@davidlmobley  
Ok I'm going to update the print parent dictionary lines then. I'll put another commit in in a minute.

bannanc:
That will print the formatted parents dictionary after each step (if verbose). 

Also - this pull request fixes a bug I didn't know we had until I was testing, which is before the trajectory file required the run to be verbose to collect data. I fixed this as well (it was just an order/tab issue in the run method)

bannanc:
@davidlmobley @camizanette pulled my original changes so she needs this fix, would it be ok to merge?

davidlmobley:
Merging!

Issue #48, opened 2016-07-06
Implement modular framework for construction of decorated chemical environment proposals


davidlmobley:
This proposes a new framework for modular creation fo more complicated and sophisticated decorators ( introduced in #13, with a first pass  in #39). We think more modularization will be needed than we have now.

Here's a recap of where we were and are:
- Prior to #39, new atom types were proposed by `&`-ing existing types with decorators, which can get us **some** of what is in parm@frosst but not everything, i.e. issues as in this comment: https://github.com/open-forcefield-group/smarty/issues/13#issuecomment-227512655 
- In #39, we have a first pass which deals with this issue by applying potentially more than one decorator at once, potentially allowing discovery of more complex types like a hydrogen bonded to a carbon that has an electron withdrawing group attached to it directly from just knowing about hydrogen (which may be necessary in AlkEthOH since there is no score benefit to finding the `HC` subtype of hydrogen from the generic if one already has `HO`; one has to find `H1` before there is a score benefit). 

We're still looking at the possible solution in #39. It works by picking a base atom type to start with, and 1-3 decorators to use. These decorators are drawn from the updated set of decorators in https://github.com/open-forcefield-group/smarty/blob/binaries/examples/AlkEtOH2/atomtypes/decorators.smarts and include bond type information, connectivity, hydrogen count, formal charge, and aliphatic or aromatic. When the chosen decorators involve a bond, such as the below, additional base atom types are chosen to fill in where the `z` character is in the bond expression: 

```
$(*-z) singly bonded
$(*=z) doubly bonded
$(*#z) triply bonded
$(*:z) aromatic bond
$(*~z) any bond
```

So, we're creating patterns which can include up to three base atom types in their SMARTS patterns, which is good as it covers substantially more chemical space. 

**However**, we think that more modularization will be needed. To explain this modularization, let's first introduce the concept of alpha and beta substituents. Given a particular atom type we are considering, an alpha substituent is its first layer of neighbors. Beta substituents are the second layer of neighbors. SMARTS pattern `[#1]` has no alpha substituent and is only a single atom. SMARTS pattern `[#1$(*-[#6])]` has an alpha substituent. SMARTS pattern  `[#1$(*-[#6]-[#8])]` has an alpha substituent and a beta substituent. Multiple beta substituents are possible, as in `[#1$(*-[#6](-[#8])-[#8])]`.

[...]

PR #59, opened 2016-07-12
Implement modular framework for construction of decorated chemical environment

camizanette:
This pull request is related with issue #48.

I am still working on this. However, since I am going to be busy tomorrow and Wednesday regarding my Advancement exam, I decided to share it like this, so people can give me feedbacks.

Here is my result after 1000 iterations:

```
INDEX        ATOMS  MOLECULES                                                          TYPE NAME                           SMARTS REF TYPE        FRACTION OF REF TYPED MOLECULES MATCHED
    1 :        244         38 |                                                       c_hydrogen                             [#1]       HC              244 /              244 (100.000%)
    2 :        232         42 |                                                         c_carbon                             [#6]       CT              232 /              232 (100.000%)
    3 :         39         30 |                                                         c_oxygen                             [#8]       OS               39 /               39 (100.000%)
    4 :         68         42 |               c_oxygen simply c_hydrogen any c_oxygenanyc_carbon      [#8$(*-[#1$(*~[#8]~[#6])])]       OH               68 /               68 (100.000%)
    5 :         68         42 |                               c_hydrogen any c_oxygenanyc_carbon               [#1$(*~[#8]~[#6])]       HO               68 /               68 (100.000%)
    6 :        152         42 |                            c_hydrogen any c_carbonsimplyc_oxygen               [#1$(*~[#6]-[#8])]       H1              116 /              116 (100.000%)
TOTAL :        803         42 |                                                                                                         767 /      803 match (95.517 %)
```

@davidlmobley, @cbayly13 and @bannanc 
I need help to find some minor errors, such as the formation of the smart strings.
I still need to treat the H2 and H3 cases (putting the parentheses on the atom types). At least it is finding H1 now!!!
I also changed the decorator file, I removed the bonds there, since we are going to add them differently, if you have another idea about how to treat this, just leave a comment.

This is just part of what is proposed on #48. As I said before, I am still working on this.

I will start to make the suggested changes on Wednesday.

Thanks!


davidlmobley:
Wow, @camizanette , this looks much improved! @bannanc , you'll want to take a look I'm sure. It's super exciting to see this finally discovering H1 for the first time -- this is really the first time seeing it learn its way to scores above 80%, but here we're seeing 97%. :) 

On this:

> I need help to find some minor errors, such as the formation of the smart strings

Do you have specific issues you'd want us to look at? Are you just talking about decoration via parentheses to get to H2 and H3, or are you talking about a problem you need help identifying? I'm sure we can help troubleshoot but it would be nice to have more guidance as to what problem you think we should look for. 

Great work!

camizanette:
It's more about if we are creating new atom types that make sense or something like that.
It would be nice to take a look at the random option I added where choose between adding a bond or a non bond decorator. Does that make sense?

About H2 and H3, I have to treat the case where we need to AddBetaDecorator with parentheses  between two bracket sets.

davidlmobley:
Yes, I'll see if I can manage to look at this tomorrow, though hopefully @bannanc can do so more easily (I'll probably be more tied up with the property calculation side of the project tomorrow with Michael Shirts).

bannanc:
Just taking a look at the code now, it already looks SOOO much better! Thanks again for your hard work @camizanette . Below are a few minor things that I think might be worth considering. 

I think bond set might be better to introduce in the init method with `self.bondset` just incase we need it elsewhere. Also @cbayly13 should we add @ to that list? It is a 'ring bond' I think it might be important when we get into bonds/angles/torsion since rings impose much different chemistry. 

~~What is natomtypes? I don't see where you define it. Is it necessary to introduce a new variable here or can we use an existing list of atoms, either `self.basetypes`, `self.atomtypes`, `self.unmatched_atomtypes`, `self.used_basetypes`, etc~~

Never mind I found this, I don't think I quite followed the new modules if they're taking number of atoms instead of an actual list of atomtypes. I'll have to look more carefully. I probably won't have time until after the outreach event and the seminar this afternoon.

Those are both very minor. I need to look more closely at the AddBetaSubstituentAtom code before I can make any suggestions here.

bannanc:
@camizanette I'm running some tests with the new code. I think I might be missing something... I keep getting errors saying unable to parse SMARTS and then SMARTS that look like this:

```
[[#8$(*~[#1$(*~[#8]-[#6])])]#[#1$(*-[#6])]]
```

Have you run into this before? I haven't investigated in the code yet, I just wanted to see if I could replicate your example...

camizanette:
@bannanc those are some of the errors I have mentioned above, I still need to treat those cases. If you find more weird SMARTS strings, please, let me know. Thank you!

bannanc:
@camizanette - I am beginning to go through the code and have a few suggestions. Would you prefer that I make changes on my branch and put in a pull request (to your branch) or just keep adding them here?

bannanc:
When we're done we will want this to be as general as possible. 

I think `PickAnAtom` should take a list of atoms. It would make more sense for call and output to be something like:

```
def PickAnAtom(self, atomlist):
...
     return atom


atom1 = PickAnAtom(self.atomtypes) 
```

that way we can easily change which list of atoms we're using. While testing for example, we could call `PickAnAtom(self.use_basetypes)` or some other list. Whenever possible methods should be general enough to use for different purposes.

jchodera:
> `[[#8$(*~[#1$(*~[#8]-[#6])])]#[#1$(*-[#6])]]`

Wow---I think the computer has surpassed my ability to understand this!

bannanc:
> Wow---I think the computer has surpassed my ability to understand this!

Yep, we're definitely getting more complex atoms we just need to get the syntax completely correct. I think the only reason that pattern can't be parsed is the extra '[' set.

davidlmobley:
This resolves #64 now, correct @camizanette ?

camizanette:
Solving issue #64 

@bannanc and @davidlmobley

davidlmobley:
From our discussion a few minutes ago, for the record, Camila also pointed out that at T=0, finding H2 and H3 is not guaranteed as it's possible to find a SMARTS which matches H1 but makes it impossible to find H2 and H3, i.e. if the SMARTS is too decorated.

davidlmobley:
OK, I skimmed the latest changes since @bannanc reviewed, and they look good to me, especially given our conversation, so I'll merge. If we want to do more code cleanup that can come in via a separate PR later.

camizanette:
It is also making weird decorations such as `[#6H2X3X3H2X3X3]` (which matches with something).

davidlmobley:
> It is also making weird decorations such as `[#6H2X3X3H2X3X3]` (which matches with something).

Can you make a separate Issue on that, giving an example? That will remind us to investigate (a) what on earth that means, and (b) how it matches anything. Show the hierarchy you showed me for that particular case...

PR #63, opened 2016-07-13
Updating Modular

bannanc:
@camizanette I've done a fair amount of rearranging of the methods you set up. 

I'm going to add comments on the "Files changed" tab so it's clear why I made the changes I did. 

The changes in the sample_atomtypes are all related to changes in other methods or changing printing to only happen when verbose.

### DIFF
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -449,6 +449,7 @@ class AtomTypeSampler(object):
             #        proposed_atomtype += bondset[0] + atom2type[0]
             #        number_brackets += 1
             proposed_typename = atom1type[1] + bondset[1] + atom2type[1]
+            if self.verbose: print("ADD FIRST BETA SUB: proposed --- %s %s" % ( str(proposed_atomtype), str(proposed_typename)))
 
         elif count > 2:
             # Has an alpha atom with at least 1 beta atom
@@ -456,11 +457,11 @@ class AtomTypeSampler(object):
             proposed_atomtype += '(' + bondset[0] + atom2type[0] + ')'
             proposed_atomtype += atom1type[0][closeAlpha+1:]
             proposed_typename = atom1type[1] + ' (' + bondset[1] + ' ' + atom2type[1] + ')'
+            if self.verbose: print("ADD MORE BETA SUB: proposed --- %s %s" % ( str(proposed_atomtype), str(proposed_typename)))
 
         else:
             # Has only 1 atom which means there isn't an alpha atom yet, add an alpha atom instead
             proposed_atomtype, proposed_typename = self.AddAlphaSubstituentAtom(atom1type, bondset, atom2type) 
-        if self.verbose: print("ADD BETA SUB: proposed --- %s %s" % ( str(proposed_atomtype), str(proposed_typename)))
         return proposed_atomtype, proposed_typename
 
 
@@ -530,13 +531,14 @@ class AtomTypeSampler(object):
                 # combinatorial-decorators
                 nbondset = len(self.bondset)
                 # Pick an atomtype
-                atom1type = self.PickAnAtom(self.atomtypes)
-                print atom1type
+                #atom1type = self.PickAnAtom(self.atomtypes)
+                atom1type = self.PickAnAtom(self.unmatched_atomtypes)
+                if self.verbose: print("atom1: %s (%s)" % ( atom1type[0], atom1type[1]))
                 # Check if we need to add an alfa or beta substituent
                 if self.HasAlpha(atom1type):
                     if self.verbose: print("has alpha")
                     bondset_index = random.randint(0, nbondset-1)
-                    atom2type = self.PickAnAtom(self.basetypes)
+                    atom2type = self.PickAnAtom(self.used_basetypes)
                     proposed_atomtype, proposed_typename = self.AddBetaSubstituentAtom(atom1type, self.bondset[bondset_index], atom2type)
                     if self.verbose: print("Attempting to create new subtype: -> '%s' (%s)" % (proposed_atomtype, proposed_typename))
                 else:
@@ -550,7 +552,7 @@ class AtomTypeSampler(object):
                         #if self.verbose: print("Attempting to create new subtype: -> '%s' (%s)" % (proposed_atomtype, proposed_typename))
                     else:
                         bondset_index = random.randint(0, nbondset-1)
-                        atom2type = self.PickAnAtom(self.basetypes)
+                        atom2type = self.PickAnAtom(self.used_basetypes)
                         proposed_atomtype, proposed_typename = self.AddAlphaSubstituentAtom(atom1type, self.bondset[bondset_index], atom2type)
                         if self.verbose: print("Attempting to create new subtype: '%s' (%s) -> '%s' (%s)" % (atom1type[0], atom1type[1], proposed_atomtype, proposed_typename))
 
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -95,6 +95,9 @@ class AtomTypeSampler(object):
         # Define internal typing tag.
         self.typetag = 'atomtype'
 
+        # Save bond list to use throughout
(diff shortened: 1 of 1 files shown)


smarty/sampler.py
```diff
@@ -95,6 +95,9 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
         # Define internal typing tag.
         self.typetag = 'atomtype'
 
+        # Save bond list to use throughout
+        self.bondset = [("-","simply"), ("=", "doubly"), ("#","triply"), (":", "aromatic"), ("~","any")]
```
bannanc, line 99: I think the bond set should be general so we can use it throughout later if we need to

smarty/sampler.py
```diff
@@ -332,38 +335,56 @@ def show_type_matches(self, atom_type_matches):
...
+    #        # Only one atom type
+    #        result = re.match('\[(.+)\]', atom1type[0])
+    #        proposed_atomtype = '[' + result.groups(1)[0] + decorators[0] + ']'
+    #        proposed_typename = atom1type[1] + ' ' + decorators[1]
+    #    return proposed_atomtype, proposed_typename
+
+
+    # I didn't understand why there were base types being introduced in this metho
+    # It should just be decorating the atom, nothing else. 
+    def AtomDecorator(self, atom1type, decorator):
```
bannanc, line 361: AtomDecorator should only add decorators, not substituents. I think this can be done more generally than we were before so I've updated this method just add a decorator to an atom with the correct syntax.

smarty/sampler.py
```diff
@@ -332,38 +335,56 @@ def show_type_matches(self, atom_type_matches):
...
         to use currently populated atomtypes, i.e. the atomtypes list, not basetypes (generic types) or 
         used_basetypes (base types which type anything)). A little silly, but still helpful. 
         Returns a tuple corresponding to the selected (smarts, name).
         """
-        atomtype_index = random.randint(0, natomtypes-1)
-        (atomtype, typename) = self.atomtypes[atomtype_index]
-        return (atomtype, typename)
-    
+        atomtype_index = random.randint(0, len(atomList)-1)
+        return atomList[atomtype_index]
```
bannanc, line 386: PickAnAtom should be able to find an atom from any atomlist, which list we use should come in the implementation part.

smarty/sampler.py
```diff
@@ -385,9 +406,10 @@ def AddAlphaSubstituentAtom(self, atom1type, bondset, atom2type):
         just be adding $(*zw) where z is the selected bond type and w is atom2type. It should raise an exception 
         if an alpha substituent is attempted to be added to an atom1type which already has an alpha substituent.
         """
-        print atom1type[0]
-        print atom2type[0]
-        print bondset[0]
+        if self.verbose:
```
bannanc, line 409: I thought we should stick with printing to the command line only when verbose

smarty/sampler.py
```diff
@@ -407,23 +429,39 @@ def AddBetaSubstituentAtom(self, atom1type, bondset, atom2type):
         an exception if a beta substituent is attempted to be added to an atom1type which does not have an alpha 
         substituent.
         """
-        s = atom1type[0][atom1type[0].find("(")+1:atom1type[0].find(")")]
-        print s
+        #s = atom1type[0][atom1type[0].find("(")+1:atom1type[0].find(")")]
+        #if self.verbose: print s
+
+        # counting '[' tells us how many atoms are in the mix
+        count = atom1type[0].count('[')
```
bannanc, line 436: I've reworked beta substituent method. I was drawing out first and second beta substitutions when I realized the beta substitution always comes after the `]` on the alpha atom. 
I think this simplifies it a little bit, but uses the framework that already existed.

smarty/sampler.py
```diff
@@ -407,23 +429,39 @@ def AddBetaSubstituentAtom(self, atom1type, bondset, atom2type):
...
+        elif count > 2:
+            # Has an alpha atom with at least 1 beta atom
+            proposed_atomtype = atom1type[0][:closeAlpha+1]
+            proposed_atomtype += '(' + bondset[0] + atom2type[0] + ')'
+            proposed_atomtype += atom1type[0][closeAlpha+1:]
             proposed_typename = atom1type[1] + ' (' + bondset[1] + ' ' + atom2type[1] + ')'
-        print "ADD BETA SUB: proposed --- " + str(proposed_atomtype) + " " + str(proposed_typename)
+            if self.verbose: print("ADD MORE BETA SUB: proposed --- %s %s" % ( str(proposed_atomtype), str(proposed_typename)))
+
+        else:
```
bannanc, line 462: If it doesn't have an alpha atom it should add an alpha atom instead, this method shouldn't be called in that situation, but I wanted a double check included.

smarty/sampler.py
```diff
@@ -491,32 +529,32 @@ def sample_atomtypes(self):
...
                 # Pick an atomtype
-                atom1type = self.PickAnAtom(natomtypes)
-                print atom1type
+                #atom1type = self.PickAnAtom(self.atomtypes)
+                atom1type = self.PickAnAtom(self.unmatched_atomtypes)
+                if self.verbose: print("atom1: %s (%s)" % ( atom1type[0], atom1type[1]))
                 # Check if we need to add an alfa or beta substituent
-                if self.HasAlpha(atom1type) == True:
-                    print "has alpha"
+                if self.HasAlpha(atom1type):
```
bannanc, line 538: Don't need to check with == True, it's already in boolean form so you can just check if it is True

bannanc:
OK, so I thought I found an H3, but it's rejecting it because it's not found in the data set... Here is the output in the log file:

```
Attempting to create new subtype:  -> '[#1$(*-[#6](~[#8])(~[#8])-[#8])]' (c_hydrogen simply c_carbonsimplyc_oxygen (any c_oxygen) (any c_oxygen))
Atom type '[#1$(*-[#6](~[#8])(~[#8])-[#8])]' (c_hydrogen simply c_carbonsimplyc_oxygen (any c_oxygen) (any c_oxygen)) unused in dataset; rejecting.
```

@davidlmobley @cbayly13 @camizanette 
am I wrong in thinking that is an acceptable SMARTS string for an H3?
Why is it `unused in dataset`?

In my most recent test it creates and rejects this at least 3 times with this exact Beta Substituent.

smarty/sampler.py
davidlmobley, line 409: Also, try for `print(` rather than `print` as `print(` works with python2.x and python3.x and `print` does not.

davidlmobley:
@bannanc - that looks right to me. Use it on the SMARTSDepict IPython notebook and confirm? It could be there's some sort of matching problem. If that's not it, it's very very very close.

smarty/sampler.py
camizanette, line 409: Those "prints" were there only for me. I was tracking everything using them, they will not be there after the work is done.

bannanc:
It finds the H3 in the substrucDepict ipython notebook [here](https://github.com/open-forcefield-group/open-forcefield-data/blob/master/Model-Systems/SMIRFF_prototype/smartsDepict.ipynb)

I'm attaching a screenshot, it's a little difficult to see, but I used this SMARTS string: 
`[#1$(*-[#6](~[#8])(~[#8])-[#8])]`
and it identifies the only H3 in the molecule with purple. I'm going to make a new issue for this.  
screen shot 2016-07-13 at 11 47 54

smarty/sampler.py
bannanc, line 409: I think they are helpful to have in general when we are tracking which atoms are being added. It might be worth leaving them for now.

camizanette:
Thank you @bannanc!!

davidlmobley:
@bannanc @camizanette - probably you already figured this out, but in case not, this was a PR into the `modular` branch, @camizanette 's branch, so merging this will:
a) automatically update the open PR into `master` to include these changes, and
b) make it so @camizanette needs to "git pull origin modular" or similar to pull down these changes into her branch

PR #71, opened 2016-07-22
Add an example of running a gas phase simulation from SMIRFF

davidlmobley:
Adding `examples/SMIRFF_simulation` which shows an example of running a single molecule simulation in vacuum beginning from a .mol2 file and the SMIRFF forcefield format.


davidlmobley:
This brings in an example of setting up a gas phase simulation of a single molecule; thanks @vtlim and @nividic. 

@nividic, @mrshirts , @bmanubay - once I merge this (momentarily) you'll find an example under examples/SMIRFF_simulation

Issue #49, opened 2016-07-06
Plan/implement chemical object infrastructure for SMIRKS-ifying SMARTY


davidlmobley:
We've realized that to go past AlkEthOH without trying to handle WAY TOO MANY special cases in construction of decorator proposals, we are going to need to move past treating SMARTS as simple strings to have the SMARTS proposal engine actually understand these. And since we have to move to SMIRKS anyway, we should really be doing this for SMIRKS and not for SMARTS.

Thus, @cbayly13 is attempting to plan the infrastructure that is going to be needed for this. 

I'll update this issue (or one of us will) as we have more details.

jchodera:
This sounds like exactly what we need. We can consider chemical graphs, where each node is annotated with SMARTS strings, and a Monte Carlo proposal process that can create and destroy connected nodes or modify the SMARTS strings on one or more nodes. [`networkx`](https://networkx.github.io) may be useful in representing these graphs since we already use it.

This might also call for an additional tool where you apply a set of SMIRKS to a huge druglike chemical dataset to see if we can type everything. We might be able to learn rules about which Monte Carlo moves are efficient (in the sense of proposing chemically reasonable SMIRKS when starting from chemically reasonable parent SMIRKS) using this.

davidlmobley:
The immediate problem we are focused on is how to even propose sensible SMARTS (or here, SMIRKS) which are suitably decorated to cover what we want, EVEN IF most of the time they don't make chemical sense. Having them mostly make chemical sense will also be an important optimization.

I'll mention the "molecules as graphs" idea to Christopher right now to see if he thinks it helps with this at all.

davidlmobley:
@jchodera : On discussion with Christopher I realized I'm confused. What are the chemical graphs you're envisioning? Are these graphs molecules, or the bipartite graph currently in Smarty where we're using atom types for graph coloring, or something else? 

Maybe this will be a topic for our catch-up call tomorrow.

jchodera:
Update: @cbayly13 will present some slides on how he thinks this might work at our Wednesday call, and then we will turn this into a detailed proposal for implementation.

bannanc:
@cbayly13 We discussed this a fair amount during the call today, I'm working on organizing my notes and writing up how we plan to implement at `ChemicalEnvironment` object which will generally use networkx graph to help organize the chemical environments we have been talking about. I'm going to work on detailed proposal and include it in a comment here when it is more organized.

davidlmobley:
[...]

PR #72, opened 2016-07-22
Add Chemical Environment Objects

bannanc:
I imagine this will still need some work, but I'm at a point that it is probably worth having people comment on my code so far. @jchodera @davidlmobley @camizanette and @cbayly13  

This is the API discussed in issue #49 where most of the details can be found.

I've tried to keep everything as general as possible. Both atoms and nodes have features that are currently OR'd together first (`bases`) then `decorators` which are AND'd with `;` at the end. the `;` is used currently because I thought the OR part should have priority. 

There are separate sub classes for Atoms, Bonds, Angles, Torsions, and Improper. I've created an ipython notebook to illustrate how to implement the objects, which is under [smarty/examples/chemicalEnvironments/](https://github.com/open-forcefield-group/smarty/tree/environment/examples/chemicalEnvironments)

I should also note that I haven't created any new tests for this section, but we will probably want to have them at some point, ideas on that front?


davidlmobley:
I ended up with way less time right now to look over this than I wanted, but a few stream-of-consciousness things which come to mind:
- `addBase` is unintuitive to me; I think it's just a documentation issue (in your notebook, in your doc strings, etc). It says it adds bases (things that are ORd together) but it doesn't help me know what IS a base, i.e. is it a SMIRKS pattern denoting a particular atom? 
- Question, does "*" have special status as a base so that it gets dropped when anything else is added, but everything else just gets ORd? 
- The IPython notebook is very helpful, though will need some editing/clarification (can provide suggestions sometime soon)
- One concrete edit is to edit the first part about adding atoms to other atoms to explain that you're elaborating the graph around the atoms involved in the selected torsion
- Make sure you have useful return values, i.e. if you try to remove an atom that can't be removed you probably want it to return something that says it did so. (Maybe it does; I do see a message printed to screen right now though.)

Functionally it's pretty nice -- I've been playing around with it a bit and it does really seem to do what we need. Nice work so far.

As soon as I have time I'll test more extensively and get to reviewing code. 

@bannanc , what's still on the "to do" list here other than documentation and tests?

Ideas on tests:
- Make sure the basic functionality you have works, i.e. run the types of things you have in your IPython notebook and check that they work
- In addition, add tests for things that shouldn't work (making sure that appropriate exceptions are raised) - i.e. what if  you try to put in a character which can't be part of a SMARTS string, or try to delete something which can't be deleted, or otherwise do things you shouldn't be doing or which would break things; the idea here is just to make sure you're raising exceptions for things that would break something if someone did them

I'll think more about all of this, but I have to dash off.

davidlmobley:
Some other things I notice doing code/functionality review, @bannanc (but overall, so far I'm very very enthusiastic as, so far, everything seems to be working properly):
- Typo in string for attempting to remove labeled atom: "Cannont remove labeled atom [*:1]"
- Should the different functionalities return a boolean indicating whether they were successful or not? i.e. `removeAtom` returns nothing, so there's no way to tell whether it successfully removed something (creating a new environment) or failed (leaving things the same). I THINK this is only an issue for removeAtom (?)
- Doc strings need some work, i.e. the main `asSMIRKS` functionality in the chemical environment class leaves me confused as to what "neighbors" would mean if provided (i.e. I don't know what the recursive calls there are referring to); it's also unclear what the point of providing `initialAtom` would be (after writing the below, I now see that this fixes which atom occurs first in the SMIRKS string, though it doesn't make the string deterministic). To some extent doc strings may benefit from usage examples. 
- Is there a problem with SMIRKS/SMARTS being non-deterministic? Right now they are not, which is somewhat surprising, but perhaps not a functional problem. For example:

``` python
bondEnv = environment.BondChemicalEnvironment()
newAtom1 = bondEnv.addAtom(bondEnv.atom1, None, None, ['#6'], None)
newAtom2 = bondEnv.addAtom(newAtom1, None, None, ['#8'], None)
print bondEnv.asSMIRKS()

# Try removing atom 1
bondEnv.removeAtom(newAtom1)

# Remove atom 2, show new SMIRKS
bondEnv.removeAtom(newAtom2)
print bondEnv.asSMIRKS()
```

gives a variety of SMIRKS back including `[*:2]~[*:1]~[#6]`, `[#6]~[*:1]~[*:2]`, `[*:1](~[#6])~[*:2]` and `[*:1](~[*:2])~[#6]`. I think there are at least six results since one can permute the order (forwards vs backwards) and also, since this can be framed as a branch point, one can also permute which atom is regarded as the branch if any). 

I'm thinking this is not a problem since all of these match the same environments with the same labeled atoms, but just wanted to bring it up. 

This **also perhaps suggests a test** -- should you loop through a complex torsion a variety of times and ensure all of the generated SMARTS/SMIRKS match the same atoms in a specified molecule? This would be a way of checking your work to go from "chemical environment" back to SMARTS/SMIRKS.

bannanc:
> Question, does "*" have special status as a base so that it gets dropped when anything else is added, but everything else just gets ORd?

"_" is a notation for wild card, but it can be modified with ands so "_;X2" is any atom that is connectivity two. So for now it is filled in when the set of `bases` (OR'd descriptors) for an atom is empty. The "*" string is not stored in `bases` there is a condition for if the set is empty. 

> One concrete edit is to edit the first part about adding atoms to other atoms to explain that you're elaborating the graph around the atoms involved in the selected torsion

@davidlmobley I don't understand what you mean by this. The `addAtom` method allows the user to specify an atom to bond to, which could be any atom in the graph object. Are you referring to the ipython notebook or the environment code?

> Should the different functionalities return a boolean indicating whether they were successful or not? i.e. removeAtom returns nothing, so there's no way to tell whether it successfully removed something (creating a new environment) or failed (leaving things the same). I THINK this is only an issue for removeAtom (?)

I like this idea, I'll think about if there are other methods which would benefit. Probably we will want something similar for `addAtom` when (if) we start using `is_valid` since there will be times you cannot add atoms because it is beyond the beta position or the atom has reached it's maximum valence, etc. 

With regard to that, I've started thinking that bonds should maybe have a feature for `order` that would recognize a '-' as a single bond and '=' as a double bond? 

> This also perhaps suggests a test -- should you loop through a complex torsion a variety of times and ensure all of the generated SMARTS/SMIRKS match the same atoms in a specified molecule? This would be a way of checking your work to go from "chemical environment" back to SMARTS/SMIRKS.

I like this idea as a test, as it shouldn't matter which atom you start with. 

> Is there a problem with SMIRKS/SMARTS being non-deterministic? 

Currently no, I think the order for neighbors in graph objects is not always the same? and if an `initialAtom` isn't specified then a random atom in the set is used. There are a few things I've been thinking about to improve this. The method has to be recursive to get the branching correct (in my opinion), but it might be better to have some kind of hidden recursive method (i.e. `_getSMIRKS`) that isn't accessible for a users. 

> Doc strings need some work, i.e. the main asSMIRKS functionality in the chemical environment class leaves me confused as to what "neighbors" would mean if provided

This is the problem I'm thinking about addressing with the hidden method, basically `neighbors` is only used in the recursive calls for the method. It addresses a problem with looping through neighbors on the graph. For example you start with atom1 and loop through it's neighbors, then you loop through atom2's neighbors, which includes atom1. In order to get the output correct we need to have a way to get atom2's neighbors without atom1. I couldn't come up with a way to do that without having another parameter in the recursive method.

bannanc:
I should also say, I'm using the key words `bases` and `decorators` for things that are OR'd and AND'd, but I would be totally willing to find and replace all of those, because the key words are not very descriptive. I thought about just calling them `OR` and `AND` or `or_descriptors` and `and_descriptors`

cbayly13:
I am still trying out the functionality; I like what I see so far. I think we need an overload of selectAtom which takes a SMIRKS-index integer as an argument, which will return the atom identified by that SMIRKS index if it exists (and presumably None otherwise). This will allow us to get specific atoms, for example the center atom of a bond angle (with selectAtom(2) ) or the one of the center atoms of a torsion.

bannanc:
> example the center atom of a bond angle (with selectAtom(2) )

I'm working on this, I'm also going to update selectBond to return the bond between two atoms if the indices are specified since that is a description we might want.

davidlmobley:
@bannanc - I think once those are in, tests are in and doc strings are updated, this will be ready to merge (of course also updating to the master branch).

bannanc:
@davidlmobley 
@cbayly13 and I had a conversation about organization tonight, I'm going to change up a few minor things that will mean updating the ipython notebook for example. With adjusting those and then the test I'll probably have the is ready tomorrow night.

bannanc:
@cbayly13 I updated the ipython notebook, you are welcome to make changes. The plan being to add a step to visualize generated structures. 

@davidlmobley I'm still working on the tests, got a bit distracted by other tasks

bannanc:
@davidlmobley I finished writing the environment object tests. Now the Travis CI build is failing. Could you help me identify what's wrong?

davidlmobley:
@bannanc - not something you broke. Could be conda updated their MD5 checksums again. Let me look.

In the meantime, you might also want to update this to master since it can't be merged until you do that anyway.

davidlmobley:
Looks like they must have changed the format of their https://repo.continuum.io/miniconda/ page again so that the code in install.sh which pulls and checks the MD5 checksum no longer works (https://github.com/open-forcefield-group/smarty/blob/master/devtools/travis-ci/install.sh). If only I were more of a `sed` wizard. Ugh. I need to update this, it appears:

```
MINICONDA=Miniconda-latest-Linux-x86_64.sh
MINICONDA_HOME=$HOME/miniconda
MINICONDA_MD5=$(curl -s http://repo.continuum.io/miniconda/ | grep -A3 $MINICONDA | sed -n '4p' | sed -n 's/ *<td>\(.*\)<\/td> */\1/p')
```

davidlmobley:
@bannanc -- see what happens if you edit `devtools/travis-ci/install.sh` to replace `MINICONDA=Miniconda-latest-Linux-x86_64.sh` with `MINICONDA=Miniconda2-latest-Linux-x86_64.sh`. It looks like they just changed the numbering. (We are already using this in some other packages, such as openmoltools, but for some reason it didn't make it in here -- i.e. see https://github.com/choderalab/openmoltools/blob/master/devtools/travis-ci/install.sh).

bannanc:
Also, when I run `nosetests -v` in my command line it doesn't print anything about the environment tests. Is there something I need to do to add the test_environment script to the tests?

Oh, it passes now. It's the same thing though, nothing showing about the environment test. I probably missed a step when making the test.

davidlmobley:
@bannanc - I've not used the class structure for tests much, but for the basic function structure, you have to have `test` in the name of the function running the test. That may be true for the class structure as well. Try?

bannanc:
Ok, yes, that was it, adding test to the function names made it work, however it's not passing. I'm investigating if it's a typo in the test script or the object themselves.

bannanc:
Ok, found a few minor typos and realized that the message printed is the first line in the function doc string so I updated those as well. 

I think they're good to go. I'll think about if there are more complicated tests we should be doing and make an issue, but at least it checks all the functions now.

PR #74, opened 2016-07-23
Adding smaller parm@frosst subset and script for creation

### LINKED ISSUE #69: Make ZINC parm@frosst subset for testing purposes
After #59 , we'll be shortly^ proceeding to the parm@frosst ZINC subset rather than AlkEthOH. @camizanette reports that running on the full parm@frosst ZINC set is a bit slow, which is of course fine once we're happy with how it's running but is not ideal for testing purposes. 

We should construct a subset (perhaps randomly) of the compounds just for testing purposes. 

^ Shortly is perhaps an overstatement since some of us have already had a look at this, but when @camizanette told me she was running some tests with parm@frosst, I said, "I feel like you just told me that the next Star Wars movie came out and you had tickets for all of us but you went without us." Now that the code is available, so the rest of us can see the show now too.

davidlmobley:
Resolve #69 by adding smaller parm@frosst subset to examples/parm@frosst/molecules; also added a script for its creation. Now instead of having 7500 molecules and taking a few seconds for every iteration, I have 500 molecules and it blasts through the whole thing rather rapidly.


davidlmobley:
Data change only, doesn't affect tests. Merging.

jchodera:
If you do end up profiling this, let me know what you find out.

For example, if the SMARTS matching is the slow part, we can take advantage of the fact that only those molecules containing atoms typed by the parent of the new proposed child need to be re-evaluated, which could speed things up considerably. As before, though, we don't want to optimize anything until we know (1) what the most time-consuming steps are, and (2) if this is really rate-limiting.

PR #73, opened 2016-07-23
Fix the "empty parents" issue of #70

davidlmobley:
Fixes the "empty parents" problem which was noted in #70 . The code which checks to ensure parents aren't emptied when children are created had gotten commented out somewhere in the process of adding construction of modular decorators.

I haven't touched the issue of why atomic decorators are redundant in #70.

PR #75, opened 2016-07-23
Fix sporadic bug introduced in #73

davidlmobley:
#73 had a lingering variable name issue which only showed up in longer tests or sporadically. Fixed here.


davidlmobley:
Also fixes a small mistake where the subset I had introduced in #74 ended up with TRIPOS atom names. This was still useful for testing (one can learn how to do TRIPOS atom typing too; we get to 88% in 1000 iterations), but was not the intended test.

Issue #14, opened 2016-06-21
Plan infrastructure for parameter ID test


davidlmobley:
Ultimately, we would like to move away from the "atom type" (or "vdw type") focus of the initial test and do a slightly more general test before moving to "real" chemistry -- particularly, we'd like to look at whether we can reproduce the **parameter types** from the parm@frosst set for AlkEthOH. This would require us to discover patterns which would reproduce vdW types as in the current `smarty`, but also reproduce bond types, angle types, and torsion types. This would test that our decoration machinery is able to handle the additional complexity that comes with that.

We'll need to think through how we would re-work/re-configure things ultimately to allow that -- probably this would basically be a step towards generating the machinery we would need for parameterizing to actual data. In this test, the existing `smarty` would be the answer to just one part of the test.

davidlmobley:
This also means we should try as much as possible to avoid the idea of "atom types" except when **scoring** the results of smarty, both because it will make this test easier and because it will make it easier for us ultimately to abandon atom types completely. (Maybe this should be a separate issue?)

davidlmobley:
This also relates to something which came up in the overly-broad issue #3 :

> > We're slowly trying to move away from smarty being just atomtypes, as the chemical perception will eventually need to handle atoms, bonds, angles, and torsions. That means reference type will be a list of atom types and atoms should be general matches. For now the scoring will likely be the fraction of reference types matches since we're still comparing to current atom types.
> > I had originally intended to write a completely different piece of code to do that, since this code has no facility for sampling over parameters, and building this in would be difficult since this capability is not modular with the current design.

I think we ought to be separating the issue of "sampling over parameters" from "handling chemical perception generally", as we ought to be able to do more than just "atom types" with smarty.

jchodera:
Do you mean that you want to try to recover bond types in parm@frosst (e.g. `HS-SH`), or the bond energetics classes?

The recovery of bond, angle, and torsion types may not be meaningful because they are all based on atom types. We're trying to get _away_ from atom types and instead discover different classes of bonds that may be divorced from atom types.

If we want to get away from atom types in a SMIRKS-based bond match, how do we test this on something that is not fundamentally atomtype based if we don't use the energetics to score?

davidlmobley:
> Do you mean that you want to try to recover bond types in parm@frosst (e.g. HS-SH), or the bond energetics classes?

[...]

PR #82, opened 2016-07-27
Add class structure/functionality to specify which parameters are used in which molecules

davidlmobley:
Adds a structure somewhat parallel to `forcefield.py` which will tell what parameter IDs/SMARTS would be used in parameterizing a specified set of molecules, for which atoms. 

Implements a `ForceField_labeler` class which is very much like `ForceField`, but the main action is performed by `labelMolecules` rather than `createSystem`. Much of the code is duplicated from `forcefield`, but no OpenMM systems are constructed/manipulated. 

The main point of this is to provide the infrastructure we need to move to a more general descendant of smarty which will use the chemical environment infrastructure of PR #72 (issue #49) to sample bond, angle, and torsional environments as well as vdW types. In this case we would be scoring based on graph coloring by "parameter ID" rather than by "atom type", i.e. as discussed in #14 .

This will also help with some other issues, such as:
- Identifying where a specific parameter is used in a set of molecules (this does exactly that): https://github.com/open-forcefield-group/open-forcefield-tools/issues/5 and https://github.com/open-forcefield-group/open-forcefield-tools/issues/4
- Identifying which parameters are used in a specific molecule (this also does exactly that): https://github.com/open-forcefield-group/open-forcefield-tools/issues/4

We'll probably need to organize the data more neatly to tabulate on a "by parameter" or "by molecule" basis to address those issues.
### To do:
- [x] Check/update doc strings and READMEs
- [x] Provide example of usage under `examples/label_molecule`
- [x] Write tests
- [x] Provide for tabulation of data to resolve  https://github.com/open-forcefield-group/open-forcefield-tools/issues/4; add test of this
- [x] Update usage example under `examples/label_molecule` of extracting detailed stats from a set of molecules


davidlmobley:
I also switched the "SMARFF" tag in the FFXML (Smarts Force Field, I assume) to SMIRFF (Smirks Force Field), as the latter is correct.

davidlmobley:
This is finished. Merging, since I also now have in place tests for all the functionality which was added here.

Issue #84, opened 2016-07-30
Extend `ForceField` class to provide infrastructure for changing parameters


davidlmobley:
As noted in #68 , we need to be able to make parameter moves. We'll initially do this by modifying the XML using `lxml.etree` which `ForceField` currently uses for XML handling.

Currently, `ForceField` only parses the XML on initialization and discards the XML after parsing. It will be updated to retain the XML. 

XML will still be parsed on initial reading of the file rather than on application of `createSystem` to avoid a speed hit when parameterizing large numbers of molecules if the XML is parsed when parameterizing each system. However, if parameters are modified, then a flag will be set so that the XML will be re-parsed when `createSystem` is applied.

jchodera:
Eventually, I think you're right that we will want to operate on a Python representation (which is created from the XML parsing) directly, and also add functions to the force generators that can reconstruct an XML subtree from their own parameters so we can write an XML from the modified parameter set. But this is a great start.

davidlmobley:
Resolved by #86 at least preliminarily.

Issue #85, opened 2016-07-30
Extra AtomTypes code in ForceField -- remove?


davidlmobley:
@jchodera - `ForceField` currently has this code snippet involved when parsing XML files:

``` python
# Load the atom masses.
        for tree in trees:
            if tree.getroot().find('AtomTypes') is not None:
                for type in tree.getroot().find('AtomTypes').findall('Type'):
                    self.registerAtomType(type.attrib)
```

As far as I can tell, there is no functionality `registerAtomType` provided in `ForceField`, so this won't work. However, we don't currently have `AtomType` definitions in the SMIRFF format, so it doesn't trigger a failure. Should it be removed?

jchodera:
I think I originally thought we needed to specify atom masses, but I must have ended up just using elemental integral masses instead (since fractional masses are dumb).

jchodera:
So I think you can get rid of this, yes.

davidlmobley:
Closing via #86

PR #86, opened 2016-07-30
Update ForceField to store XML internally, add very basic parameter moves

davidlmobley:
This will update `ForceField` to retain the XML internally and allow its modification (i.e. for parameter moves). The XML will still be parsed on `ForceField` initialization, as per #84 , but if it is modified, it will trigger re-parsing and storing of forces on application of `createSystem`. 

I expect I'll implement some basic functionality here for interacting with and manipulating parameters, such as:
- `GetParameter`, which would return info associated with a specific parameter as identified either by SMIRKS or parameter ID and force type
- `SetParameter`, mirroring `GetParameter` but for setting a particular parameter
- `GetParameterID`, returning just the parameter ID for a particular parameter
- `GetParameterIDs`, getting all parameter IDs for a particular section (needed for determining unique parameter IDs for new child parameters)

Presumably later we will also need the ability to introduce new parameters, such as something like:
- `CreateChildParameter(smirks, parentsmirks, section, paramdict)`, 
- `DeleteChildParameter` which would delete a specified parameter which is a child of another parameter

Along with that latter set, we would probably also need something which tracks the parameter hierarchy so we know which parameters can be deleted (at least without substantial work, only parameters which have no children can be deleted).


jchodera:
Looks good to me.

davidlmobley:
Updated to remove unused code which attempted to handle `AtomTypes` (which we're not using) as per #85 .

davidlmobley:
Added `getParameter`, `setParameter`, and `writeFile`.

davidlmobley:
Updated to add a (detailed) test and an example of parameter manipulation in an IPython notebook under `examples/forcefield_modification`. 

Should be ready to merge as this has the basics (get a parameter, set a parameter, write XML). Other functionality will be discussed in a separate issue and added under another PR.

@bannanc - take a quick look and try my IPython notebook under `examples` if you don't mind...?

smarty/forcefield.py
```diff
@@ -389,9 +386,144 @@ def registerGenerator(self, generator):
...
+
+Usage notes: SMIRKS or parameter ID must be specified.
+
+To do: Update behavior of "Implied" force_type so it raises an exception if the parameter is not uniquely identified by the provided info.
+"""
+        # Check for valid input
+        if smirks and paramID:
+            raise ValueError("Error: Specify SMIRKS OR parameter ID but not both.")
+        if smirks==None and paramID==None:
+            raise ValueError("Error: Must specify SMIRKS and/or parameter ID.")
```
bannanc: You raise an error if both smirks and paramID are specified, doesn't that mean the Error message here should say "SMIRKS or parameter..." not "and/or"?

smarty/forcefield.py
```diff
@@ -389,9 +386,144 @@ def registerGenerator(self, generator):
...
```
bannanc: "and/or" --> "or"?

bannanc: "and/or" to "or"?

bannanc:
@davidlmobley it took me a little bit to try to follow this since I haven't been working on this side of the project. I made comments on the code lines, but there were a few Error messages that I think need to be or not and/or. That is assuming I understood your if statements  correctly. 

Otherwise, I think I followed the ipython notebook and it looks good.

davidlmobley:
Thanks, @bannanc .

PR #89, opened 2016-08-01
Make asSMIRKS methods deterministic and more user friendly

bannanc:
We discussed this briefly in the first environment pull request, but I never made an issue. It was a faster fix than I thought so I just created a pull request instead. 

@davidlmobley here is your initial comment:

> Doc strings need some work, i.e. the main asSMIRKS functionality in the chemical environment class leaves me confused as to what "neighbors" would mean if provided (i.e. I don't know what the recursive calls there are referring to); it's also unclear what the point of providing initialAtom would be (after writing the below, I now see that this fixes which atom occurs first in the SMIRKS string, though it doesn't make the string deterministic). To some extent doc strings may benefit from usage examples.

I have updated things with the asSMIRKS methods a bit. First, I adjusted the internal working so that the printed SMIRKS from asSMIRKS always prints the same string. While different orders would identify the same fragment, I think it is preferable for the string to be deterministic. I also created an internal method (_asSMIRKS) that does the recursive method strategy so the user will not see the `initialAtom` or `neighbors` parameters which primarily existed for the recursive use. 

I had a method asSMARTS in the Atom class that creates a string for a single atom in the form smarty uses now. For example:`[#1$(*-[#6])]` instead of `[#1]-[#6]` I've renamed this method asAtomtypeSMARTS so it's more clear what it is doing. 

@cbayly13 I don't think this would change anything you're working on, but heads up for the asAtomtypeSMARTS name change.

### DIFF
--- a/examples/chemicalEnvironments/test_environments.ipynb
+++ b/examples/chemicalEnvironments/test_environments.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 20,
    "metadata": {
     "collapsed": false
    },
@@ -27,7 +27,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": 21,
    "metadata": {
     "collapsed": false,
     "scrolled": true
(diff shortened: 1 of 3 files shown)


bannanc:
If I realized how fast this change would be to make I would have included it in my last pull request.

davidlmobley:
LGTM.

PR #91, opened 2016-08-02
Create getters and setters for ORtypes and ANDtypes

bannanc:
Per @cbayly13 's request there are now four new methods for Atoms and Bond objects

`getORtypes`
`setORtypes`
`getANDtypes`
`setANDtypes`

That allows the user to get, change, and set  OR or AND types for atoms and bonds. 

@davidlmobley These methods are pretty basic, but it would let us change the underlying structure and still let users interact with it in the same/similar way.


davidlmobley:
Add a basic test of these, @bannanc ?

davidlmobley:
Merging now, adding test later.

Issue #103, opened 2016-08-06
Support for bond charge corrections


jchodera:
It's very easy to add support for forcefield-defined bond charge corrections, so I'm going to do that now, since it will enable dielectric corrections (timely with @cfennell's visit).
### API Proposal

I propose to extend the API of `ForceField.createSystem()` to add a `chargeMethod` argument, which can take one of the following values:
- `'BCC'` will apply bond charge corrections if the `ffxml` file had a `<BondChargeCorrections>` block. This first applies a semiempirical method (specified in the `method` attribute of the `<BondChargeCorrections>` XML block) and then applies bond charge corrections. If no `<BondChargeCorrections>` section is specified, the charges specified in the `OEMol` `molecules` are used.  **(default)**
- Any of the charge methods in `openeye.oequacpac` can be specified as a string (e.g. `'OECharges_AM1BCCSym'`), and this charge model will be used.
- `None` forces the charges from the provided `OEMol` `molecules` argument to be used. (useful for testing)
### XML specification

Rules:
- Later definitions that match the _same bond_ override earlier definitions. This is needed to ensure the definitions scheme is hierarchical, like other parts of SMARTY.
- Multiple definitions can be sequentially applied to the _same atom_.

``` XML
<BondChargeCorrections method="AM1" increment_unit="elementary_charge">
  <BondChargeCorrection smirks="[#6X4:1]-[#6X3a:2]" increment="+0.0073"/> <!-- tetrahedral carbon bonded to aromatic carbon correction -->
  <BondChargeCorrection smirks="[#6X4:1]-[#6X3a:2]-[#7]" increment="-0.0943"/> <!-- tetrahedral carbon bonded to aromatic carbon (bonded to a nitrogen) -->
</BondChargeCorrections>
```

where the `"AM1"` in `method` would be expanded to `OECharges_AM1` for `oechem.oequacpac` in our initial implementation.
### Example usage

Current scheme remains default: Take charges from `molecules`:

``` python
system = forcefield.createSystem(topology, molecules)
```

Use `ffxml`-defined bond charge corrections:

``` python
system = forcefield.createSystem(topology, molecules, chargeMethod='BCC')
```

Use recommended AM1-BCC from `oequacpac`:

``` python
system = forcefield.createSystem(topology, molecules, chargeMethod='OECharges_AM1BCCSym')
```

**Update:** Proposal tweaked so that any bond charge corrections in the file are automatically applied.

jchodera:
[...]

PR #104, opened 2016-08-06
ForceField bond charge correction and charge handling

jchodera:
This implements #103 (Support for bond charge corrections).

I also discovered (and fixed) a bug in the nosetests that utilized generators that prevented most of the `ForceField` tests from actually running.
## 
### API

The API of `ForceField.createSystem()` has been extended to add a `chargeMethod` argument, which can take one of the following values:
- `'BCC'` will apply bond charge corrections if the `ffxml` file had a `<BondChargeCorrections>` block. This first applies a semiempirical method (specified in the `method` attribute of the `<BondChargeCorrections>` XML block) and then applies bond charge corrections. If no `<BondChargeCorrections>` section is specified, the charges specified in the `OEMol` `molecules` are used.  **(default)**
- Any of the charge methods in `openeye.oequacpac` can be specified as a string (e.g. `'OECharges_AM1BCCSym'`), and this charge model will be used.
- `None` forces the charges from the provided `OEMol` `molecules` argument to be used. (useful for testing)
### XML specification

Rules:
- Later definitions that match the _same bond_ override earlier definitions. This is needed to ensure the definitions scheme is hierarchical, like other parts of SMARTY.
- Multiple definitions can be sequentially applied to the _same atom_.

``` XML
<BondChargeCorrections method="AM1" increment_unit="elementary_charge">
  <BondChargeCorrection smirks="[#6X4:1]-[#6X3a:2]" increment="+0.0073"/> <!-- tetrahedral carbon bonded to aromatic carbon correction -->
  <BondChargeCorrection smirks="[#6X4:1]-[#6X3a:2]-[#7]" increment="-0.0943"/> <!-- tetrahedral carbon bonded to aromatic carbon (bonded to a nitrogen) -->
</BondChargeCorrections>
```

where the `"AM1"` in `method` would be expanded to `OECharges_AM1` for `oechem.oequacpac` in our initial implementation.
### Example usage

Current scheme remains default: Take charges from `molecules`:

``` python
system = forcefield.createSystem(topology, molecules)
```

Use `ffxml`-defined bond charge corrections:

``` python
system = forcefield.createSystem(topology, molecules, chargeMethod='BCC')
```

Use recommended AM1-BCC from `oequacpac`:

``` python
system = forcefield.createSystem(topology, molecules, chargeMethod='OECharges_AM1BCCSym')
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,12 @@ script:
   # Add omnia channel
   - conda config --add channels ${ORGNAME}
   # build the recipe
-  #- conda build devtools/conda-recipe
+  - conda build devtools/conda-recipe
   # Test the local installation
-  #- source activate _test
+  - source activate _test
   # This is a temporary workaround for the conda issues.
   - conda install --yes nose nose-timer numpy networkx lxml openmm
-  - pip install .
+  #- pip install .
   # Install OpenEye toolkit
   - pip install $OPENEYE_CHANNEL openeye-toolkits
   # Run tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
   - pip install $OPENEYE_CHANNEL openeye-toolkits
   # Run tests
   #- cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=2 --with-doctest --with-timer -a "\!slow" && cd ..
+  - python --version
   - cd devtools && nosetests -vv --nocapture --with-timer $PACKAGENAME && cd ..
 
 env:
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -24,6 +24,8 @@ test:
   requires:
     - nose
     - nose-timer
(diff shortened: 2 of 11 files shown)


jchodera:
Note that this doesn't yet apply best-practices conformational expansion (ELF) explicitly, unless this is handled transparently "under the hood" by `OEAssignPartialCharges`, but this can be added in a separate PR by @cbayly13. Charge assignment is encapsulated in the function `_assignPartialCharges()` to make it easier for you to modify.

We can also later add caching of partial charges to speed up repeated calls to `createSystem()`.

jchodera:
Fixes #107

jchodera:
Finally fixed all of the py3 issues!

jchodera:
I need the functionality and py3 bugfixes in this PR to make progress on https://github.com/open-forcefield-group/open-forcefield-tools/pull/14, so I'm going to merge this now. Please comment on #103 if you think we should make changes to how the charging API works, and we can fix things in another PR.

PR #113, opened 2016-08-12
Merge forcefield "labeling" code into forcefield.py

### LINKED ISSUE #105: Merge "labeling" code from forcefield_labeler back into ForceField class now that XML is retained
It appears that a bunch of code from `forcefield.py` was just copied into `forcefield_labeler.py` and modified a bit. This creates a huge technical debt nightmare.

We can't keep this. We need to do one of the following:
- Someone needs to take complete responsibility is it to keep the code in `forcefield_labeler.py`  completely current with `forcefield.py`.
- We integrate its functionality into `forcefield.py` in a minimally invasive way.
- We delete `forcefield_labeler.py` if it no longer serves a purpose.

Otherwise, the technical debt of maintaining two nearly-identical pieces of code is enormous, and we can't spare the effort on this.

davidlmobley:
Resolves #105 by allowing `ForceField` to provide information on which parameters will be used for specified oemols. Removes `forcefield_labeler.py` and now all the requisite functionality is within `forcefield.py` and especially in the `ForceField` class.

Usage is now, for example:

``` python
from smarty.forcefield import *
ff = ForceField( get_data_filename('forcefield/Frosst_AlkEtOH.ffxml') )
labels= ff.labelMolecules( [mol], verbose = True )
```

rather than:

``` python
from smarty.forcefield_labeler import *
labeler = ForceField_labeler( get_data_filename('forcefield/Frosst_AlkEtOH.ffxml') )
labels= labeler.labelMolecules( [mol], verbose = True )
```

@bannanc and @bmanubay , take particular note as I believe you've been using the old functionality a bit.

Tests were updated accordingly, and forcefield_labeler.py is now completely removed, as are its tests.


davidlmobley:
The tests will likely fail because of OpenEye license expiration, though they pass locally.

jchodera:
`_getSMIRKSMatches_OEMol` is identical to `getSMIRKSMatches` except for essentially a single line. Can you reconcile these?

Maybe you can expose a public `getSMIRKSMatches_OEMol()` helper function, and simply use this in `ForceField.getSMIRKSMatches()`?

smarty/forcefield.py
```diff
@@ -671,11 +717,66 @@ def createSystem(self, topology, molecules, nonbondedMethod=NoCutoff, nonbondedC
...
+
+        # Storage for labels
+        molecule_labels = []
+
+        # Loop over molecules and label
+        for idx,mol in enumerate(oemols):
+            molecule_labels.append({})
+            for force in self._forces:
+                # Initialize dictionary storage for this force type
+                if isinstance(force, HarmonicBondGenerator):
```
jchodera: This violates the principle that each force plugin should not require changes to the main `ForceField` code. Can you get the label from the force generator key instead?

jchodera: For example, you could just use `force.__class__.__name__` if you are OK with `HarmonicBondGenerator`, or you can add a class field to each force generator like `label = 'HarmonicBondForce'`.

davidlmobley:
> _getSMIRKSMatches_OEMol is identical to getSMIRKSMatches except for essentially a single line. Can you reconcile these?

No. getSMIRKSMatches operates on a Topology file and unrolls all instances matching a particular pattern. _getSMIRKSMatches_OEMol operates on an OEMol and does not unroll all matches. The "operate on" difference is helpful because we want to be able to do things like ask whether a given molecule contains a particular parameter before we bother creating a topology, etc. And we don't unroll, because there's no need to find out that all instances of the same molecule contain the same parameters. 

I could make it public, however.

jchodera:
You know, I bet this whole thing could be simplified enormously if you did two things:
- Compile the molecule list into a `Topology` that contains one copy of each molecule
- Compile your lists of `force_terms` if a `self.label` field is set in the `createForce()` call for each force generator. This would prevent you from creating separate `parseForce()` functions for each generator.

That should eliminate most of the code!

jchodera:
> No. getSMIRKSMatches operates on a Topology file and unrolls all instances matching a particular pattern. _getSMIRKSMatches_OEMol operates on an OEMol and does not unroll all matches. 

We could have two functions:
- Operate on an OEMol and return matches for that molecule
- Operate on a [topology, molecules] and unroll matches

The second function could call the first to eliminate code duplication.

davidlmobley:
OK. I think these are all good ideas. Functionally it should work now, but I can get to these tomorrow or Monday I hope. 

Things for me to do:
- [x] Expose public `getSMIRKSMatches_OEMol` functionality to operate on oemol and return matches, possibly renaming
- [x] Update existing `getSMIRKSMatches` (renaming to unrollSMIRKSMatches, since it no longer does the matching itself) to operate on a topology and oemols and unroll matches; this will call the function above to eliminate code duplication
- [x] Use class names to avoid hardcoded force labels in `ForceField` 
- [x] Rename `parseForce` to `labelForce` to make more clear this only has to do with labeling matches

davidlmobley:
Relating to this, @jchodera :

> You know, I bet this whole thing could be simplified enormously if you did two things:
> 
> Compile the molecule list into a Topology that contains one copy of each molecule
> Compile your lists of force_terms if a self.label field is set in the createForce() call for each force generator. This would prevent you from creating separate parseForce() functions for each generator.
> That should eliminate most of the code!

I'm actually missing how this dramatically simplifies things, and it seems to add overhead that I don't think we want. Specifically:

1) **Simplification**: It would definitely eliminate `parseForce`, but there is very little duplication between `createForce` and `parseForce` code, so if I make it so `createForce` can retrieve the info that `parseForce` currently does, it will make `createForce` necessarily longer.
2) **Overhead**:  `createForce` requires an OpenMM system, whereas `parseForce` does not. This means that if I want to, say, query a large set of molecules to see what parameters they used, I will have to construct an OpenMM system for each of the molecules first, whereas now I do not. 

Overhead may not be a big deal now, but down the road I can see us using this functionality to, say, query a large set of molecules from, say, eMolecules to see what's available that provides information about a particular parameter or would use that parameter (and how many occurrences are in said molecules). It seems silly to have to create an OpenMM system for each to be able to do that inquiry, and I imagine would slow the process significantly.

I wrote it the way I did precisely to avoid having to create OpenMM systems for simple inquiries like this, as in fact you suggested in an earlier call when we discussed this idea.

jchodera:
> I'm actually missing how this dramatically simplifies things, and it seems to add overhead that I don't think we want. 

The similarity between `createForce` and `parseForce` may be only superficial then. Let's punt on that change until later until we can figure out if it would actually simplify things.

What about changing `parseForce` to `labelForce` to make it clear that this has only to do with labeling matches?

davidlmobley:
> What about changing parseForce to labelForce to make it clear that this has only to do with labeling matches?

Absolutely, will do.

davidlmobley:
Addressed the above issues and especially removed the redundancy between `getSMIRKSMatches_OEMol` and `getSMIRKSMatches` by renaming the latter to `unrollSMIRKSMatches` and it now uses the former to do the actual matching.

I verified that correct energies are still produced for the AlkEthOH test after these changes. 

Passes tests locally and is, I believe, ready to merge.

davidlmobley:
Tests fail on Travis because of OpenEye issues, but I'll merge since it looks good locally.

jchodera:
Thanks!

PR #115, opened 2016-08-13
Implement conformer generation prior to charging via recommended procedure

### LINKED ISSUE #114: Implement better partial charge assignment via best practices
I'll be implementing as described here: https://github.com/open-forcefield-group/smarty/issues/103#issuecomment-239499877

davidlmobley:
Resolves #114 by adding conformer generation prior to charging to `_assignPartialCharges` in `forcefield.py`. See also #103 .


davidlmobley:
This passes tests now. I could merge, though it would perhaps be good if I actually wrote out an instance of a charged molecule and inspected it just to make sure nothing is obviously awry first.

jchodera:
Sounds great.

davidlmobley:
I've verified that this does not do anything crazy. I've also verified that it does give significantly different charges for a case where I might expect a modest difference, 1,2-ethanediol. So, as far as I can tell this is working, and I will merge once tests pass.

davidlmobley:
This is passing tests since I didn't change anything since it was. Travis lists it as failing, but it appears travis is getting stuck while fetching things. I'll merge since I need to pull this into another open PR.

PR #117, opened 2016-08-22
Bring Bayly's latest envMoves examples from his branch into environment branch, updating existing PR #112

bannanc:
I would like to have @cbayly13 's example ipython notebooks in the environment branch. I think everything they are testing fits within the current WIP pull request #112


davidlmobley:
Thanks, sounds good. I just changed the name of this to describe what it actually does.

(Names might not seem important right now but they can be very helpful when looking back at a history of changes, which is sometimes necessary.)

PR #126, opened 2016-08-26
Adding utility function C Bayly needs for inspecting problem parameters.

davidlmobley:
Adds a utility function to `forcefield_utils`, `getMolParamIDToAtomIndex`, which constructs a dictionary Christopher Bayly needs. The function takes an oemol and constructs a dictionary, keyed by parameter ID, of the corresponding SMIRKS and then a list of all of the atom indices to which it is applied.


davidlmobley:
Merging as C Bayly needs this. Will add Travis tests of this functionality in a later PR.

Issue #125, opened 2016-08-25
Implement MDL aromaticity model option for handling SMIRFFs


davidlmobley:
In Christopher Bayly's parm@frosst SMIRFF draft, he's going to be relying on an alternate aromaticity model. Initially we thought he would need to kekulize all of the structures but it turns out that won't do what he needs, so instead we'll be using the MDL aromaticity model. 

It turns out that OEPrepareSearch clobbers whatever aromaticity model is used in the underlying OEMol, so we will need to add an option to SMIRFF to use this alternate aromaticity model and then bypass OEPrepareSearch by preparing our own search. Christopher has provided some example code for this.

``` python
mol = oechem.OEMol()
unique = True
for mol in ifs.GetOEMols():
    #print( mol.GetTitle(), mol.NumAtoms())
    oechem.OEClearAromaticFlags( mol)
    oechem.OEAssignAromaticFlags( mol, oechem.OEAroModel_MDL)
    # avoid running OEPrepareSearch or we lose MDL aromaticity, so instead:
    oechem.OEAssignHybridization( mol)
    oechem.OEAssignFormalCharges( mol)
    oechem.OEAssignImplicitHydrogens( mol)
    for triple in subschToATypes:
        subsch = triple[0]
        #oechem.OEPrepareSearch(mol, subsch)
        for match in subsch.Match(mol, unique):
            #print( 'have a match with', triple[2])
            #display( depictMatch(mol, match))
            for ma in match.GetAtoms():
                #print(ma.target.GetIdx(), triple[1], end=" ")
                ma.target.SetType( triple[1])
            #print('')
    oechem.OEWriteConstMolecule( ofs, mol)
    for atom in mol.GetAtoms():
        ofsTypes.write( '%s %s\n' % (mol.GetTitle(), atom.GetType()) )
```

This would be in lieu of the `kekulize` option proposed in https://github.com/open-forcefield-group/smarty/issues/53; probably we'd trigger via `aromaticitymodel=MDL` or something along those lines. Will come up with a concrete proposal when i get ready to implement (which will need to be soon).

davidlmobley:
Ah, actually, it looks like we are not currently using `OEPrepareSearch` in `ForceField`; the search is done via whatever aromaticity model/preparation is applied to the provided OEMol. This means an interim solution for @cbayly13 is to apply the desired aromaticity model prior to providing the oemol(s) to `ForceField`.

However, I should still implement something to invoke this. In #53 I'd proposed:

[...]

PR #127, opened 2016-08-26
Add aromaticity model support as per 125

davidlmobley:
This adds aromaticity model support as per #125 . Also adds a version attribute at the same time.

C Bayly needs this ASAP so I'm trying to get something rough up immediately for him to use. I'll revisit it likely early next week to polish.


davidlmobley:
We've verified that this works. On `c12c(cccc1)occ2` the bond 6-7 (among others) ends up different with and without the MDL aromaticity model in Chris's "super preliminary draft parm@frosst-y SMIRFF for bonds".

But, I have to run off for the weekend, so Bayly will be working/testing.

jchodera:
Please add nosetests! "we've verified that this works" is not future proof!

davidlmobley:
@jchodera - yes, will add nosetests. That's why this is still [WIP]. I've just been focused on getting things done which must be done now before implementing tests. :)

davidlmobley:
Added tests verifying not just that it runs but also that the choice of aromaticity model has the desired effect. Should be ready to merge once it passes tests.

Issue #53, opened 2016-07-07
Add fractional bond orders to SMIRFF


davidlmobley:
Eventually, we need to add capacity to SMIRFF for fractional bond orders, which we will obtain from a semi-empirical QM calculation. This will be used downstream eventually (after this summer) for parameter interpolation.

jchodera:
Suggested FFXML format:

``` XML
<HarmonicBondForce length_unit="angstroms" k_unit="kilocalories_per_mole/angstrom**2" fractional_bond_order="interpolate-linear">
   <Bond smirks="[#6:1]-[#6:2]" length1="1.526" k1="620.0" length2="1.426" k2="820.0" length3="1.326" k3="1020.0"/> 
   ...
</HarmonicBondForce>
```

davidlmobley:
So, an interesting question is whether we would allow fractional bond orders only for SMIRKS matching `~` bond types or whether we would allow them for any. Presumably it would be for any, I suppose, as you have here. Otherwise the concern would be that the semiempirical calculation would give us a fractional bond order but the previous SMIRKS would prevent us from using it... Is that what you're thinking?

@cbayly13 , does this look suitable?

cbayly13:
We want to exclude triple bonds so the form I would suggest is:
[#6:1]!#[#6:2]
which is asking for a carbon-carbon bond of any order except triple, thus allowing double, single, and aromatic. Most useful conjugation is across CC, CN, and CO bonds so I would want to start with
[#6:1]!#[#6:2]
[#6:1]!#[#7:2]
[#6:1]~[#8:2]
and see how far we can get with that. In the last case only a "~" is needed because singlet state C#O does not occur in pharmaceutically relevant molecules.

cbayly13:
In principle triple bonds would not be ruled out, because they can conjugate too, but I am expecting a lot of finessing needed to map: 
single<-->conjugatedSingle<-->aromatic<-->conjugatedDouble<-->double 
correctly, which is where all atom-typed ligand force fields fall down.

davidlmobley:
@jchodera - for reasons I can explain in more detail when I see you in person later this week, we may now want to do this sooner rather than later (in brief, it has to do with trying to put parm@frosst in SMIRFF format, which will be MUCH MUCH easier for specific types of systems if we have this). 

[...]

PR #134, opened 2016-09-02
Implement partial bond orders

davidlmobley:
Begins implementation of tracking bond orders (and partial bond orders, when desired) as per #53 . This PR will cover handling of partial bond orders for bond parameters only. The influence of partial bond orders on other parameters (i.e. angles, torsions) will come later in a separate PR.

To do:
- [x] Update `_Topology` to include `self._bondorder` obtained when reference molecules are processed
- [x] Create example XML file which shows updates for partial bond order (placed example in `examples/partial_bondorder`. Currently this only uses partial bond orders to collapse a bond parameter, though there is a torsion which this could also be applied to.
- [x] Update charge calculation to update bond orders to store Wiberg bond orders to molecules, if present
- [x] Because the user will not always calculate charges, update the `_assignPartialCharges` method so it can be used store partial bond orders if the user is not doing a charge calculation; this will involve doing an AM1 calculation but not modifying charges.
- [x] Add a `_Topology` method, `_updateBondOrders(self)` to re-populate a topology's bond orders (`self._bondorders`) with bond orders for the molecules involved. This can take an optional argument, `Wiberg = True`, which, if present, will use Wiberg bond orders; if not present, it will use the original bond orders. In this way it can take over for the present `_initializeBondOrders(self)` (introduced above) by handling its functionality AND updates based on Wiberg.
- [x] Update `createSystem` to ensure partial bond orders are assigned (either via charge calculation or `_assignPartialCharges` with `modifycharges = False`)
- [x] Update bond parameter assignment to use partial bond orders when assigning bond parameters (when the FFXML asks for it, of course)
- [x] Add tests (including both where the user asks for a charge calculation and where the user does not)


davidlmobley:
I'll need to test this on something where this will allow collapsing parameters. Please note the following from Christopher Bayly's smirff99Frosst branch from a SMIRKS-style modified frcmod file for `[#6X3]~[#6X3]` bonds:

```
[#6X3:1]-[#6X3:2]                410.  1.45    parm99 generic
[#6X3:1]:[#6X3:2]               469.0  1.40    parm99 generic
[#6X3:1]=[#6X3:2]               549.0  1.35    parm99 generic
```

These can collapse into a single parameter using interpolation if we format them as `[#6X3]~[#6X3]`, so this should be a good initial test case. 

In SMIRFF format these are:

``` XML
...
    <Bond smirks="[#6X3:1]-[#6X3:2]" id="b5" k="820.0" length="1.45"/>
    <Bond smirks="[#6X3:1]:[#6X3:2]" id="b6" k="938.0" length="1.40"/>
    <Bond smirks="[#6X3:1]=[#6X3:2]" id="b7" k="1098.0" length="1.35"/>
...
```

This should become something like:

``` XML
<HarmonicBondForce length_unit="angstroms" k_unit="kilocalories_per_mole/angstrom**2" fractional_bondorder="interpolate-linear">
...
    <Bond smirks="[#6X3:1]!#[#6X3:2]" id="b5" k1="820.0" k2="1098" length1="1.45" length2="1.35"/>
...
```

I didn't make these numbers up - it turns out the lengths scale perfectly linearly if we think of aromatic as a bond order of 1.5, and the force constants are very close to linear (linear would be 959 rather than 938).

There is a side issue we'll revisit at some point, though -- the Wiberg bond orders seem to run a little low because of "leakiness", so Chris thinks we may prefer sigmoidal to linear so that things that are 1.0-1.2 or so tend to end up as single bonds and things which are 1.8-1.9 tend to end up as doubles, etc. We'll probably have to investigate a bit. But I'll finish implementing linear and then look into this.

jchodera:
@davidlmobley : You can merge or rebase from master since #135 fixed travis tests.

davidlmobley:
Ah, right, this is failing tests because this feature requires OpenEye October 2016 release, which is now in beta. I need to update the installation/requirements to get this...

davidlmobley:
This passes tests now and I will merge. 

Please note that this switches Smarty to use (and require) the OpenEye October release toolkits which are currently in beta ( @bannanc @camizanette @bmanubay ). these can be obtained via:
`pip install --pre -i https://pypi.anaconda.org/openeye/label/beta/simple openeye-toolkits --upgrade`
for example.

The change is needed as partial bond orders are not available in earlier toolkit versions.

PR #136, opened 2016-09-07
Add prototype of smirff99Frosst.ffxml

davidlmobley:
Per Chris Bayly, this adds "smirff99Frosst.ffxml, a small-molecule smirff force field based on parm99.dat and parm@Frosst. It has specific parameters for most molecules in the parm@Frosst Zinc subset of 7505 molecules (17 exceptions for torsions, most of those with broken chemistry). More testing to follow."

This is basically the product of his final couple weeks of work (which will conclude Thursday). It's neither parm99 nor parm@frosst exactly, for a number of reasons including that:
- It covers, or should cover with only slight modification, all reasonable organic chemistry, which neither  of the above did
- It is much simpler (because of parameter consolidation by SMIRKS, because of removing questionably differentiated parameters, and because its goal is to be a "good starting point" than a finished product)
- It introduces a number of approximations/consolidations for conciseness/clarity
- It fixes a number of mistakes/omissions, etc. 

Presumably Christopher will elaborate if his schedule permits.


jchodera:
Awesome!

Some thoughts:
- We can keep the prototype here for automated testing, but can we migrate
  the "canonical" forcefield home to `open-forcefield-group/smirff99frosst`
  or similar? We can use this as an opportunity to work out some of the
  concepts of how we would manage releases of production forcefields from our
  Bayesian scheme.
- A separate repo would allow us to have official GitHub release versions
  of the forcefield, which will make it very useful for referring to specific
  versions of the forcefield.
- It's totally OK if the first few GitHub release versions are
  "prereleases" and called something like 0.1 or `2016-09-07-dev` or
  something to indicate their prerelease status.
- If we enable Zenodo for this repository, we will get citable DOIs for
  each release.

We'll want to figure out how we can use this forcefield alongside existing
protein and nucleic acid forcefields at some point soon. Is it expected
that we would still use ParmEd for this purpose, or would we want to do
something to allow the existing OpenMM ForceField class (which can assign
AMBER parameters) to work alongside the SMARTY ForceField class?

bannanc:
I just pulled this down to look at how this does typing molecules in DrugBank. 

~~I'm continually getting an error when importing convert_frcmod. This line:
 `from convert_frcmod import *`
Was definitely too early in the sequence as it came before smarty was imported.~~

Edit:
 I found the problem, there is an ipython notebook in examples at `smarty/examples/smirff99Frosst` and another in utilities at `smarty/utilities/convert_frosst/` We should probably only have one or the other to reduce confusion. It also appears that the version in examples is the more updated version from @cbayly13 with visualization of the fragment in question.

davidlmobley:
@jchodera :

> - We can keep the prototype here for automated testing, but can we migrate
>   the "canonical" forcefield home to `open-forcefield-group/smirff99frosst`
>   or similar? We can use this as an opportunity to work out some of the
>   concepts of how we would manage releases of production forcefields from our
>   Bayesian scheme.

I think this sounds like a good idea. @bannanc -- do you see any problem with us setting this up? I'd set it up. 

Presumably the repo would have different folders for different types of releases, e.g. releases from the Bayesian scheme, releases based on manual creation (such as this), releases based on fitting with fixed chemical perception (such as if Lee-Ping wants to run ForceBalance down the line to fit a forcefield, etc). 

The point about using Zenodo to get DOIs for the releases is also a good one.

jchodera:
I think we want one repo per forcefield. That way the DOI applies to a
single release of a single forcefield.

bannanc:
I agree with John on the multiple repos, if the point of the separate repository is to track new versions and make the forcefield citable, it makes more sense to treat each of these separately. 

I'd be happy to get a smirff99Frosst repo going. Personally, I would rather do the testing and incremental improvements over there so there is less confusion between the smarty/smirky updates and the updates to smirff99Frosst.

davidlmobley:
I'll try and start a separate forcefield repo for smirff99Frosst today, @bannanc . And yes, I agree about a repo per forcefield.

davidlmobley:
@bannanc - now that the smirky branch is merged in, would it makes sense to update this one to master and get it merged in? (Or is there another branch that's got this in it and which should be brought in instead?)

Once this is merged in, we can separately deal with the "permanent home" of force fields. Though, I'll go try and create the relevant repo right now.

bannanc:
I have the test_smirff99Frosst branch that is still rather unorganized, but has changes not included here. I will merge into this and get it organized so that everything related to the smirff99Frosst is in the same place in the repository, there are currently 3 copies of the ipython notebook for example. 

It shouldn't take too long, I'll get it done today.

davidlmobley:
Thanks, @bannanc ! Once it is in, we can put the actual FFXML file (and, probably, the source files from Christopher since these are important!) in the repo, which is now up: https://github.com/open-forcefield-group/smirff99Frosst

bannanc:
I'm working on cleaning up the directories for smirff99Frosst and the associated tests. 

I think it makes the most sense to keep these all in utilities/convert_frosst since they aren't really examples of how to use anything it is just the process of testing the ffxml to see what molecules use any given parameter. 

Christopher had this statement in the original README file at [examples/smirff99Frosst](https://github.com/open-forcefield-group/smarty/tree/smirff99Frosst/examples/smirff99Frosst) "Only a few of the files here are intended for use, most are intermediate files used to generate the final useful ones. The intermediate ones are here in case a forensic audit is needed for certain parameters."

there are 16 intermediate smirffishFrcmod files, I'm certain we don't need to keep ALL of them, but some might be useful to understand the process of converting the original frcmods to a the smirffish ones...

davidlmobley:
@bannanc -- until he comes along to tell us which we wouldn't ever want to look at (or unless you're clear on that) I'd just keep all of them. I do know some of the ones I might want to look at, but I also imagine that in certain cases I might just end up grepping them to find what I was looking for... 

(What I imagine is that if we find issues with a parameter, we will want to say, "well, where did THAT come from" and so we'll go poking back into the intermediate files here, many of which are commented to indicate how he got to where he ended up...)

davidlmobley:
(Note that this has some conflicts, presently, on the stuff you recently changed in smarty, so you'll need to fix before we merge it.)

bannanc:
Yeah, I'm not ready to merge quite yet. I'm going to create an "intermediate_smirffish" directory to keep those files and then we have them as needed. 

I'll pull master again, when I get some of this sorted.

bannanc:
Assuming we want to merge the smirff99Frosst that is still a work in progress this is ready to go. I still have some of parameters that are assigned generics on zinc and the drug bank sets. 

I include the script I've been using to filter molecule sets, I added some commentary and examples for it. Someone @davidlmobley or @camizanette probably, or I can bring in Vickie if you guys don't have time.

utilities/convert_frosst/convert_frcmod.py
```diff
@@ -90,6 +90,7 @@ def convert_frcmod_to_ffxml( infile, inxml, outxml ):
     force_section = ['NonbondedForce', 'HarmonicBondForce', 'HarmonicAngleForce', 'PeriodicTorsionForce', 'PeriodicTorsionForce']
     ct = 0
     thissec = None
+    # Why is this a while loop and not a for line in text loop?
```
davidlmobley, line 93: No good reason, other than that because I was going to be skipping lines with comments, coding this way came to mind first rather than using `for` with `continue` statements.

davidlmobley (commented):
This looks good to me, @bannanc . I'll go ahead and merge.

utilities/filter_molecule_sets/remove_smirks.smarts
```diff
@@ -0,0 +1,4 @@
+% This is an example of SMIRKS that would be removed
+% one column of smirks patterns like:
+[#103:1]
+% There are no Lawerencium atoms in any set so this won't affect the current filtering
```
davidlmobley, line 4: 👍

Issue #122, opened 2016-08-24
Potential issue with impropers in ForceField


davidlmobley:
Quite by accident, I discovered that if I have an FFXML file which contains a generic improper such as
`<Improper smirks="[*:1]~[*:2]~[*:3]~[*:4]" id="i0" k1="1000" periodicity1="2" phase1="180."/>`

then no proper torsions get assigned to my molecules, only the above improper.

This may not be a major issue since most "realistic" impropers wouldn't actually match patterns that torsions would match, I think. BUT, it probably also means there is a bug in handling of torsions/impropers in `ForceField`.

davidlmobley:
OK, on further investigating/thinking about it, I don't think this is actually a bug, but really a "user error". Specifically, my SMIRKS pattern is incorrect for an improper. Our convention in SMIRFF is that the central atom is the second atom, so an improper would be formed as `[*:1]~[*:2](~[*:3])~[*:4]`, NOT `[*:1]~[*:2]~[*:3]~[*:4]`, as the latter matches torsions. If it occurs in a SMIRFF file following the propers, all propers will be overwritten by the generic. 

So, probably I need to implement "improper checking" in SMIRFF to ensure that all impropers are properly formed with the second atom as the central atom. I'll have to bug @bannanc to see if she knows the easiest way to do this.

davidlmobley:
Somewhat relatedly, there is a note in the code asking whether we need to handle propers and impropers differently. I think the answer is "no" except that impropers need to be checked to see if they are well formed, to avoid gobbling up propers. https://github.com/open-forcefield-group/smarty/blob/master/smarty/forcefield.py#L1392

davidlmobley:
I think the fix (for safety) to the issue here is to have `periodicTorsionGenerator.createForce` do a separate check for impropers to ensure they are correctly formed (the second atom is central). Otherwise, sooner or later, someone is likely to accidentally omit a branch in an improper and have it gobble up a bunch of their torsions.

bannanc:
I have a built in check with environment objects that if you initiate them with a SMIRKS string the type should match the object you initiated. So if you create a `ImproperChemicalEnvironment` with a SMIRKS that's not for an improper it should raise a SMIRKSMisMatchError, example below:

```
import smarty.environment as env
smirks = "[*:1]~[*:2]~[*:3]~[*:4]" 
improper = env.ImproperChemicalEnvironment(smirks)

[...]

Issue #51, opened 2016-07-07
Add suitable keyword for Smirff trefoil SMARTS impropers


davidlmobley:
SMIRFF will use improper torsions which involve a trefoil-type SMIRKS pattern. The plan is to apply these to all (three? six? Bayly thinks 6) paths which constitute an improper to avoid problems with ordering/ambiguity.

We need to add a suitable keyword for specifying these to SMIRFF.

Not high priority yet.

jchodera:
How about a `symmetrize="True"` keyword to symmetrize the impropers?

davidlmobley:
Sounds perfect to me.

cbayly13:
Do we need a keyword beyond simply the parameter type of "Improper"? When would we not be symmetrizing?
BTW there are 6 impropers around a trefoil.

jchodera:
> Do we need a keyword beyond simply the parameter type of "Improper"? When would we not be symmetrizing? BTW there are 6 impropers around a trefoil.

I don't believe the current behavior of `tleap` includes all 6 impropers around a trefoil, so if we want to match the behavior of other codes for validation or reproducibility reasons, we would need a way to select the behavior mode.

I'm all in favor of having the default mode be the "good" behavior, but providing a compatibility mode (e.g. `mode="amber"`) for backward-compatibility with other codes.

davidlmobley:
Yes, I totally agree with @jchodera here. We don't want to put ourselves in a world where we couldn't implement something similar to AMBER, CHARMM, etc. as a SMIRFF if we wanted to. 

Specifically, if we made it so all impropers were this way it seems possible that it it might make it impossible to reproduce parm@frosst energies on the full ZINC set. So we should retain the possibility of having old-style behavior if and when we want it.

cbayly13:
_Sigh_ I suppose we do have to reproduce the old flawed way of doing things.

jchodera:
OK, which behavior do we want?
- Default to adding all 6 trefoil impropers, obtain old behavior with a keyword (which one?)
- Default to adding one improper, add all 6 trefoil impropers if a keyword is specified (which one?)
- Require keyword to be specified so behavior is unambiguous

Note that even if we add just one improper, we probably won't add the same improper that `tleap` is adding, so as @cbayly13 has pointed out before, we have to be careful to separate impropers in the energy component validation checks because these will not be the same.

davidlmobley:
[...]

Issue #62, opened 2016-07-12
Validate energies/forces for impropers - start with benzene


davidlmobley:
The AlkEthOH set contains no impropers, so we will need to test against those eventually once (a) #51 is implemented and (b) we expand out to include things that have impropers, such as adding phenyl.

A good test would be o-xylene.

davidlmobley:
@cbayly13 will shortly push a SMIRFF to his branch which covers all of the impropers from parm@frosst and parm99 (which ends up being only four terms once all the atom types are collapsed down as SMIRKS, if I have it right), so this is basically ready to be tested. The place to start is benzene.

davidlmobley:
Resolved by #143 .

PR #143, opened 2016-09-21
Validate/finalize implementation of impropers

davidlmobley:
This currently adds an IPython notebook looking at energies for benzene and o-xylene from parm@frosst and smirff99Frosst to validate the implementation of impropers (as per #62). It also provides (in `examples/testing/test_impropers`) an example modified smirff99Frosst which has been tweaked to have the same angle energies within the ring as parm99 for benzene (at the cost of an extra angle parameter) to aid with energy validation.

Before merged, this will also:
- [x] Divide six-fold (trefoil) impropers by six internally so that AMBER barrier heights can transfer in directly, as per https://github.com/open-forcefield-group/smarty/issues/51
- [x] Add automated testing of the improper implementation (i.e. on benzene) along with a minimal benzene ffxml file to enable this
- [x] Add check that impropers are impropers and propers are propers (#122)
- [x] Also test on molecule `CC(C)(C)c1ccccc1C(C)(C)C` which uses steric strain to induce significant improper energy (benzene and o-xylene don't).

I also fixed a small bug with an improper smirks provided in one of the tests (it was not being checked previously).


davidlmobley:
This now resolves #51 in that the six-fold division is now taken care of internally.

davidlmobley:
Resolves #122 by checking that no SMIRKS patterns for impropers are labeled as propers or vise versa.

davidlmobley:
@bannanc -- can you try running the SMIRKS for propers/impropers through your chemical environments using the latest AlkEthOH ffxml? This PR is throwing an error in Travis because it can't recognize whether one of the torsions (proper or improper) is a proper or an improper. It could be that there is a typo, or it could be a problem with your environments.

The error is something like:

```
if OR[0] != '~':
IndexError: string index out of range
```

and it's thrown by `test_forcefield.py` on Travis.

The relevant code is in forcefield.py in this PR around line 1363 and is based on the code snippet you sent me, i.e.:

```
            # Check that the SMIRKS pattern matches the type it's supposed
            # to be (avoiding bugs wherein an improperly formed generic improper
            # overrides propers, for example)
            chemenv = env.ChemicalEnvironment(self.smirks)
            thistype = chemenv.getType()
            if thistype=='Torsion': thistype = 'Proper'
            if self.torsiontype != thistype:
                raise Exception("Error: SMIRKS pattern %s (parameter %s) does not specify a %s torsion, but it is supposed to." % (self.smirks, self.pid, self.torsiontype))
```

but apparently it's having trouble recognizing one or more of the SMIRKS patterns used in AlkEthOH.

bannanc:
@davidlmobley  
Without looking at the details I can bet where the problem is. Currently my parsing expects SMIRKS input the same as environments output. That is there are no "embedded atoms" so it doesn't handle SMIRKS in the form `[#1$(*~[#6])]` yet. It just looks for the beginning and ends of atoms based on where the `']'` symbol is. 

I'm working on a plan to include this in our parsing, but I haven't incorporated it yet.

davidlmobley:
Merging this, as now impropers are validated and functional, and testing is in place.

Issue #133, opened 2016-09-02
Create documentation of SMIRFF FFXML format


davidlmobley:
I'll need to create documentation of the SMIRFF FFXML format on here, as it needs to be documented other than in the `ForceField` help.

davidlmobley:
Draft documentation is up now. Comments/suggestions welcomed.  https://github.com/open-forcefield-group/smarty/wiki/The-SMIRFF-force-field-format

PR #147, opened 2016-09-22
Update SMIRKS parsing for chemical environments; modify forcefield to monitor SMIRKS parsing errors

bannanc:
@davidlmobley I thought I would get you the improved smirks parsing first to your branch so you could implement your smirks checks. 
I updated the test_environment script and fixed a typo in the cli_smirky script. 
I also updated the AlkEtOH ffxml file to replace element letters with SMARTS [#...] patterns. There weren't many of them. I'm going to take a look at the "smirffish" file that we converted into smirff99Frosst.

I need to add more comments/documentation to the new functions in environment, that change is coming soon, but I thought I would get this up so you could see it.


davidlmobley:
Thanks, I'll have a look! I went ahead with the immediate issue without this by just bypassing the check whenever your environment couldn't parse the SMIRKS (and printing a warning) but I can try this out and see if it makes a big difference in the number of warnings.

bannanc:
@davidlmobley Most recent commit includes better comments and I did a little reorganization. It raises a SMIRKSParsingError if the parsing function doesn't work so at least we have a specific exception to look for.

bannanc:
~~Is there a way to edit a pull request to change the base? There's not a reason to go into the issue62 branch if that is already merged?~~

Since issue62 was merged into the master, I switched the base for this pull request.

davidlmobley:
There might be, @bannanc , but I'm not sure. I don't think it's needed -- I'll update my code to look for the exception you implemented anyway, so I'll test this out after you're done and I merge it into issue62, then do a new PR into main which will also include my changes.

bannanc:
Ok, you can change it with edit, then you can switch the branch base you're comparing to. 

~~I switched it back to issue62, I guess we should get the new parser to you first to test how it is working.~~ 

Keeping the merge to master and communicating with David to merge in changes from issue63 branch

utilities/convert_frosst/smirffishFrcmod.parm99Frosst.txt
```diff
@@ -135,18 +135,18 @@ ANGL
 [*:1]~[#6X4:2]-[*:3]      50.  109.5   parm99 generic Csp3
 [#1:1]-[#6X4:2]-[#1:3]    35.  109.5   parm99 generic Csp3
 # Csp3 in 3-membered ring
-[*r3:1]1~@[*r3:2]~@[*r3:3]1 350.  60.     Frosst generic any atom within the 3-membered ring
-[*r3:1]~@[*r3:2]~!@[*:3] 100.  118.    Frosst generic 3ringAtom-3ringAtom-substit
-[*:1]~!@[*r3:2]~!@[*:3]  100.  113.    Frosst generic substit-3ringAtom-nonHsubstit
-[#1:1]-[*r3:2]~!@[*:3]    50.  113.    Frosst generic substit-3ringAtom-Hsubstit
+[*;r3:1]1~;@[*;r3:2]~;@[*;r3:3]1 350.  60.     Frosst generic any atom within the 3-membered ring
+[*;r3:1]~;@[*;r3:2]~;!@[*:3] 100.  118.    Frosst generic 3ringAtom-3ringAtom-substit
+[*:1]~;!@[*;r3:2]~;!@[*:3]  100.  113.    Frosst generic substit-3ringAtom-nonHsubstit
```
davidlmobley, line 140: Was the old one a valid smirks here?

davidlmobley:
@bannanc - tests are failing now...?

(Unrelatedly -- were any of the changes to smirff99Frosst actually bugs, or all just changes for easier parsing?)

davidlmobley:
Ahh, right, I see -- tests are failing because you made this a PR into master and it doesn't include the adjustments I need to make to `forcefield` yet, @bannanc . Let's talk for a minute to sort out your plans, as i could resolve this several ways:
a) Leave this as a PR into master and modify your branch or do a PR into your branch to fix the issue with `forcefield`
b) Make this a PR into my `issue62` branch, merge it, and then fix the issue
etc.

davidlmobley:
NOTE for the record: I made and pushed to this branch an update to `smarty.forcefield` to use the new exception this raises in order to better check for problems with SMIRKS parsing. This also makes it so the tests pass. 

@bannanc - pull down updates before proceeding, please.

utilities/convert_frosst/smirffishFrcmod.parm99Frosst.txt
bannanc: So the record is here, these were valid before, it is a personal preference that makes the parsed smirks into environment objects make more sense.

bannanc:
~~I forgot that there was an issue with the improper in alkethoh not getting parsed as an improper... I'm going to explore further, but the updated SMIRKS in AlkEtOH and smirff99Frosst can be parsed into environments, I haven't checked types yet. 

Following up, it doesn't seem to be a parsing issue, the smirks pattern `"([a,A:1]~[#6X3:2]([a,A:3])~[#8X1:4])"` isn't parseable into environments for some reason... I'll have to look into this further, I'll try to sort it out this weekend.~~

Ignore this, I had an interactive python script still open from last night that didn't have update scripts in it. 

@davidlmobley I think everything is working now, the SMIRKS `"([a,A:1]~[#6X3:2]([a,A:3])~[#8X1:4])"` parses and is typed as an `Improper` correctly.

jchodera:
Would it make sense to add a check for incorrectly formatted impropers? It should be easy to add something to `ForceField` that checks that the bond connectivity pattern of the `Topology` has all atoms connected to atom number 3.

davidlmobley:
@jchodera - yes, exactly, we're headed in that direction. The first step was an update (in #143) to check that propers are propers and impropers are impropers.

That's actually what this is discussing. We're now using the chemical environments to check that torsions (propers and impropers, the latter being the more challenging) are formatted properly. HOWEVER, some of them are challenging to parse properly, which is what this was about.  Caitlin had to make a variety of enhancements to the chemical environment objects so that they could understand the full array of SMIRKS Christopher had come up with.  

@bannanc -- when/if we get to the point where the environments can understand everything in smirff99Frosst (is that the case now?) I should update `forcefield` to actually insist that all of the torsion smirks parse properly into chemical environments, rather than just doing what I implemented in #143 and having it check whether they are the correct type WHEN they parse properly.

(For @jchodera -- right now, we attempt to parse the smirks into a chemical environment, and then if it parses, we check that it's the correct type; if it doesn't parse, we just raise a warning and keep going. This means everything which actually parses into the environments as a bond, angle, torsion, or vdW SMIRKS will result in an Exception if it's not categorized correctly, but anything which doesn't will result in a warning. It's a major effort to make sure the environment objects can understand the entire SMIRKS language or at least all of it that's relevant here...)

jchodera:
I was suggesting something much simpler than relying on valid `ChemicalEnvironment` parsing: You can add a very quick check to the force generators to make sure that atoms 1 and 2 in a `HarmonicBondForce` are connected with a bond in the `Topology`; that there is a bond between 1-2 and 2-3 in the `Topology` for a `HarmonicAngleForce`, etc. This is probably just a few lines of code per force generator and will catch simple mistakes or weird discrepancies (especially in impropers).

davidlmobley:
@jchodera - that's a good idea. I hadn't thought of waiting until the smirks get turned into a force and checking via the Topology. I'll see if I can implement that in the near future.

davidlmobley:
Actually, @jchodera -- now that I think about your idea a little more, I think it will only provide partial checking, and only for angles, torsions and impropers. Specifically, for bonds, a SMIRKS pattern is parsed and used to create a bond, then the smirks pattern is not retained in a way which is tied to the bond. So, as long as bond creation worked, we WILL end up with a bond between the atoms... 

This will be slightly more useful for angles, torsions and impropers, as we can check the bonding pattern for the atoms involved in the force to make sure it is correct.

jchodera:
Maybe I still haven't communicated the idea clearly. Let me create a PR so
you can see what I mean---it's only a few lines.

davidlmobley:
For some reason it wasn't clicking earlier, but now it is. Good idea.  And I'll check out the pr at some point, thanks!

jchodera:
Looks like it's already caught some bugs:
https://travis-ci.org/open-forcefield-group/smarty/jobs/162512374#L1921-L1937

davidlmobley:
@jchodera -- I'm confused about this one, as we're using trefoil impropers of this form in SMIRFF:
`<Improper smirks="[*:1]~[#6X3:2](~[*:3])~[*:4]" id="i1" k1="1.1" periodicity1="2" phase1="180."/>`

(This is the one your test is erroring on, for benzene. If you're not used to reading this, it's saying that it should look for any atom with a single bond to a trivalent carbon (which is the central atom) then a branch that can be any atom, and the trivalent carbon is also connected to another atom -- i.e. a trefoil with the trivalent carbon (atom 2 if you count from 1) in the center.

Here, if we number atoms from 0, then atom 1 is the central atom. Your test is giving this error:

```
assert topology._isBonded(atom_indices[0], atom_indices[2]), 'Atom indices %d and %d are not bonded in topology' % (atom_indices[0], atom_indices[2])
```

So it's looking to see if atom_indices[0] is bonded to atom_indices[2], which it shouldn't be.

jchodera:
Whoops. I was under the impression that the third atom was central. Is it
always the second atom in our convention?

davidlmobley:
Yes, I think, @jchodera , your assertion for impropers is treating `atom_indices[2]` as if it's the central atom, whereas it should be`atom_indices[1]` with the trefoil impropers we are supposed to be using.

(I need to do the documentation on SMIRFF as per #133 and this is something which should go in there.)

FWIW, I don't think there IS any other way to write trefoil impropers sensibly with SMIRKS, as you have to do a branch off of the central atom, which necessitates that the central atom be atom 2. That is to say, one could naively attempt to write a trefoil a couple of possible ways by just permuting where the branch is:
1. `[*:1]~[*:2](~[*:3])~[*:4]` 
2. `[*:1](~[*:2])~[*:3]~[*:4]`
3. `[*:1]~[*:2]~[*:3](~[*:4])`
but only option 1 actually works, since option 2 has the branch off of the first atom (which is not what you want) and option 3 doesn't work either.

davidlmobley:
Yes, @jchodera , second atom. Apologies -- this is something which has been drawn on the board many times, and is obvious if you look at the SMIRKS, but it's not documented anywhere yet. I really need to get to #133. I have a whole list of things which need to go into it somewhere around...

davidlmobley:
(Side note -- the reasons the tests picked this up is that you're looking at the test I just put in to test whether the benzene improper is working properly. While I wouldn't be surprised if what you're putting in revealed bugs in impropers, it wouldn't be there, as that one already (a) gives correct energies as compared to AMBER, and (b) is correctly parsed by the chemical environments into a chemical graph for an improper. In other words, it's already passing much more stringent tests...)

jchodera:
Looks like the rule that impropers have all bonds going to the second atom [isn't correct either](https://travis-ci.org/open-forcefield-group/smarty/jobs/162519571#L1937):

```
FAIL: Test implement of impropers on benzene.
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/smarty/tests/test_forcefield.py", line 413, in test_improper
    g0, g1, e0, e1 = compare_molecule_energies( top, crd, ff, mol, skip_assert = True)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/smarty/forcefield_utils.py", line 262, in compare_molecule_energies
    smirfftop, smirffsys, smirffpos = create_system_from_molecule(forcefield, mol, verbose = verbose)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/smarty/forcefield_utils.py", line 99, in create_system_from_molecule
    system = forcefield.createSystem(topology, [mol], verbose=verbose)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/smarty/forcefield.py", line 940, in createSystem
    force.createForce(system, topology, verbose=verbose, **kwargs)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/smarty/forcefield.py", line 1487, in createForce
    assert topology._isBonded(atom_indices[0], atom_indices[1]), 'Atom indices %d and %d are not bonded in topology' % (atom_indices[0], atom_indices[1])
AssertionError: Atom indices 3 and 11 are not bonded in topology
```

What's the convention here, then?

bannanc:
I'm trying to catch up on this, but I'll answer @davidlmobley 's question for me first

> @bannanc -- when/if we get to the point where the environments can understand everything in smirff99Frosst (is that the case now?) I should update forcefield to actually insist that all of the torsion smirks parse properly into chemical environments, rather than just doing what I implemented in #143 and having it check whether they are the correct type WHEN they parse properly.

I added a line to [convert_frcmod.py](https://github.com/open-forcefield-group/smarty/blob/environment/utilities/convert_frosst/convert_frcmod.py) to test that the SMIRKS coming from the "smirffish" file can be parsed into environments. All of the smirff99Frosst SMIRKS and AlkEtOH_Frosst.ffxml SMIRKS are also parsed into environments correctly. I must have forgotten to note that change here.

davidlmobley:
The convention I stated is correct,  @jchodera, so I'll have to investigate in detail.  This is passing the chemical environment test  so the smirks should be correct, so I don't see (without checking code and files which I can't do right now)  how this error could arise.  You're not doing a check which is order dependent (1-0 but not 0-1), are you?  Anyway,  if Caitlin can't investigate ill get back to this tonight.

jchodera:
The bonds should be added in a symmetric manner:
https://github.com/open-forcefield-group/smarty/pull/148/files?diff=unified#diff-09e736461d2b8108fc74549e8902230bR255

davidlmobley:
Well, @jchodera , nothing immediately is striking me in thinking about this, so I'm going to have to dig in in detail in the next couple of days. The improper in question really does have the bonding pattern I'm talking about -- `[*:1]~[#6X3:2](~[*:3])~[*:4]` -- so, there must be a bookkeeping error somewhere or other, either in your code or mine. Maybe a torsion is getting miscategorized as an improper, or somehow some atom numbering is getting botched somewhere, or something. I'll probably have to drill down and print out the atom indices that the SMIRKS is matching, check that they are correct against the .mol2 file, and so on. Anyway, thanks for the help -- these tests will clearly be helpful to have there; it's just puzzling what's going wrong right now.

davidlmobley:
@jchodera -- @bannanc seems to have figured out what's going on here. It looks like the culprit is `ValenceDict`, which reorders atoms when the first element is larger than the last element. Since impropers are order-dependent, this must not be applied for impropers, as it will result in an incorrect central atom whenever an improper is reordered.

I can submit a PR to your branch (and PR) to fix this -- I think it should just be a matter of making the torsion a simple dictionary when it is an improper, and a `ValenceDict` when it is a normal torsion.

jchodera:
> @jchodera -- @bannanc seems to have figured out what's going on here. It looks like the culprit is ValenceDict, which reorders atoms when the first element is larger than the last element. Since impropers are order-dependent, this must not be applied for impropers, as it will result in an incorrect central atom whenever an improper is reordered.

Great catch!

> I can submit a PR to your branch (and PR) to fix this -- I think it should just be a matter of making the torsion a simple dictionary when it is an improper, and a ValenceDict when it is a normal torsion.

Is that correct? If you are using trefoil impropers and the second atom is the central atom, any of these orderings would be equivalent:

```
1-2-3-4
3-2-1-4
4-2-1-3
3-2-1-4
...
```

Don't you need a specialized version to handle these permutations?

davidlmobley:
@jchodera -- we already get all six matches we want out of the SMIRKS matching. You probably don't remember but the decision we made in a previous call was that we want to separately apply all six of these, i.e. instead of each improper parameter resulting in one force it results in six. We then divide the original barrier height by six so that we still end up with ``the same'' force we otherwise would have had (avoiding the crazy order-dependencies that AMBER had).

Or am I misunderstanding what you're saying?

davidlmobley:
Note to self: I was overlooking the fact that the dictionary is used by all torsions, not just on a per-torsion basis, which makes this more complicated... 

Not sure the easiest path to resolve this. I'm thinking... (Suggestions appreciated.)

jchodera:
Huh. I can't actually find the place where it applies this to all impropers around the trefoil in [`forcefield.py`](https://github.com/open-forcefield-group/smarty/blob/master/smarty/forcefield.py). All I see is that you [divide `k` by 6](https://github.com/open-forcefield-group/smarty/blob/master/smarty/forcefield.py#L1388-L1391).

davidlmobley:
@jchodera -- this happens because the impropers match all six paths.

davidlmobley:
That is to say, we ask for a `ss.Match( mol, unique)` with `unique=False`.

davidlmobley:
(I've verified that this results in the force getting applied to all six paths.)

davidlmobley:
Hmm, but you may be right -- it looks like Christopher on occasion will add some decoration in one of the positions in an improper ASIDE from the central position, which I think would result in fewer SMIRKS matches so we'd have to expand these back out, i.e. see: https://github.com/open-forcefield-group/smirff99Frosst/blob/master/smirff99Frosst.ffxml#L122 where there's this SMIRKS: `[*:1]~[#6X3:2](~[#8X1:3])~[#8:4]` which would I believe result in one or two matches depending on the structure of the molecule in question.

I was testing only on SMIRKS of this form `[*:1]~[#6X3:2](~[*:3])~[*:4]` (and had seen only smirks of that form for impropers) when I wrote this; these always result in six matches so no expansion was needed. 

Thoughts on the easiest way to expand out the other terms around the trefoil? I'm guessing I need something which expands all the equivalent paths after I do the smirks matching.

So really there are two separate issues:
- [ ] Ensure impropers don't get reordered in the same way propers do (instead, they should be reordered in a different way)
- [ ] Add all impropers around a trefoil by making sure we cover all six permutations even if the SMIRKS pattern doesn't give all of them

jchodera:
Ah, OK! Maybe we should be clearer about that in documenting the code.

What about other force terms that match the same sets of atoms multiple times? I guess this is eliminated by the symmetries enforced by `ValenceDict`?

We could separate code paths for impropers and propers by changing [this code](https://github.com/open-forcefield-group/smarty/blob/master/smarty/forcefield.py#L1412-L1416) to `registerProperTorsion` (which is symmetric) and `registerImproperTorsion` (which uses a normal `dict()`), and splitting the subsequent code paths.

davidlmobley:
> What about other force terms that match the same sets of atoms multiple times? 

We had to add this (switching from `unique = True` to `unique = False`) otherwise you miss torsions in cyclobutane, for example, where a `1-2-3-4` torsion and a `4-3-2-1` torsion are not the same as a `2-3-4-1` torsion. :) If `unique = True`, the substructure search only gives you a single torsion for cyclobutane... 

But, `unique = False` is set for unrolling SMIRKS matches in every case. 

> I guess this is eliminated by the symmetries enforced by ValenceDict? 

Yes. The ONLY thing affected in our tests (in AlkEthOH) by switching from `unique = True` to `unique = False` was the set of torsions we ended up with for four-membered rings. (Still to be done: Check three membered rings, where related issues will arise.)

> Ah, OK! Maybe we should be clearer about that in documenting the code.

It IS documented in the code doing the substructure searhc.

davidlmobley:
> We could separate code paths for impropers and propers by changing this code to registerProperTorsion (which is symmetric) and registerImproperTorsion (which uses a normal dict()), and splitting the subsequent code paths.

Ah, excellent, that sounds like a good solution. Let's do that. (Shall I, or do you want to?)

PR #149, opened 2016-09-27
Add plots for results to smirky

bannanc:
I have added the data plots to that @camizanette and @davidlmobley have been using for smarty to smirky. I also updated the `load_trajectory` function to use pandas so it is more general. That means that the parsing works even when there are commas in the SMIRKS patterns. Not a lot of changes, but I thought I would make it available. I'm working on a few other updates to odds in smirky, but I'm not quite done yet.

### DIFF
--- a/smarty/cli_smirky.py
+++ b/smarty/cli_smirky.py
@@ -8,6 +8,7 @@ import string
 import time
 
 from optparse import OptionParser # For parsing of command line arguments
+import smarty
 
 import os
 import math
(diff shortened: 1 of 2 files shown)


jchodera:
Cool! Any chance you could post an example plot?

davidlmobley:
LGTM, I'll merge. Do post a plot for John when you have a chance, @bannanc . John, they're basic at this point -- primarily score versus time if I remember correctly.

bannanc:
@jchodera I left a longer simulation going when I left the office tonight, I'll post the resulting plot in the morning.

PR #150, opened 2016-09-27
Adjust SMIRFF and fractional bond orders to only use when desired

davidlmobley:
#134 implemented partial bond orders for handling of bond parameters (as per #53) via Wiberg bond orders available via AM1 calculations in the October OpenEye beta toolkits. However, my initial implementation required that Wiberg bond orders always be available when `createSystem` is applied to a molecule, whether or not the forcefield would actually use them. This is bad for at least a couple of reasons:
1. It requires all users to have the October OpenEye toolkits
2. It makes system setup unnecessarily slow (because it requires an AM1 calculation) when user provided charges are desired and no fractional bond orders are needed

This adds an optional `use_fractional_bondorder` attribute of the SMIRFF tag which must be set to true if fractional bond orders are going to be used in the specified forcefield XML file (otherwise an exception will be raised when they are encountered). If this tag is not set, then no AM1 calculations will be done to obtain fractional bond orders (e.g. if the user supplies charges).

I'll need to get this into the documentation of #133.

@camizanette -- you might see if this will help with your issue.


davidlmobley:
@camizanette - when you have a chance, please see if this allows the forcefield test to run locally on your computer with the non-beta version of openeye; if it gets that working for you, then I will merge.

davidlmobley:
I just caught a small bug wherein the default charge method for `createSystem` was set to `BCC` rather than None (contrary to the doc string), and fixed it. Also, the doc string for `createSystem` referred to the `chargeModel` argument rather than the `chargeMethod` argument so I fixed this as well.

jchodera:
Thanks!

davidlmobley:
I just verified that these changes make it so I can use `ForceField` again with the older OpenEye tools as long as I don't try to use partial bond orders, so it should be good to merge once tests pass.

davidlmobley:
Python3.4 got stuck for some reason but the others passed, so I'm merging.

Issue #16, opened 2016-06-21
Possible bug when specialized basetypes are used


cbayly13:
To try and start with the highest score in smarty, I tried the AlkEthOH parm@Frosst smart used to generate the reference atom types... this should be a perfect match:

```
% atom types
[$([#1]-C)]   HC
[$([#1]-C-[#7,#8,F,#16,Cl,Br])]   H1
[$([#1]-C(-[#7,#8,F,#16,Cl,Br])-[#7,#8,F,#16,Cl,Br])]   H2
[$([#1]-C(-[#7,#8,F,#16,Cl,Br])(-[#7,#8,F,#16,Cl,Br])-[#7,#8,F,#16,Cl,Br])]   H3
[#1$(*-[#8])]   HO
[#6X4]   CT
[#8X2]   OS
[OX2+0$(*-[#1])]   OH
```

These were used at the basetypes in a smarty run for which the log shows correct matches (under ATOMS):

```
INDEX        ATOMS  MOLECULES                                                          TYPE NAME                           SMARTS
    1 :        244         38 |                                                               HC                      [$([#1]-C)]
    2 :        116         37 |                                                               H1  [$([#1]-C-[#7,#8,F,#16,Cl,Br])]
    3 :         33         22 |                                                               H2 [$([#1]-C(-[#7,#8,F,#16,Cl,Br])-[#7,#8,F,#16,Cl,Br])]
    4 :          3          3 |                                                               H3 [$([#1]-C(-[#7,#8,F,#16,Cl,Br])(-[#7,#8,F,#16,Cl,Br])-[#7,#8,F,#16,Cl,Br])]
    5 :         68         42 |                                                               HO                    [#1$(*-[#8])]
    6 :        232         42 |                                                               CT                           [#6X4]
    7 :         39         30 |                                                               OS                           [#8X2]
    8 :         68         42 |                                                               OH                 [OX2+0$(*-[#1])]
TOTAL :        803         42
```

but no matches to reference atom types:

```
Atom type matches:
HC                                                               matches       H3 :        0 atoms matched
H1                                                               matches       HO :        0 atoms matched
H2                                                               matches       OS :        0 atoms matched
H3                                                               matches       HC :        0 atoms matched
HO                                                               matches       H1 :        0 atoms matched
CT                                                               matches       OH :        0 atoms matched
OS                                                               matches       H2 :        0 atoms matched
OH                                                               matches       CT :        0 atoms matched
[...]

PR #151, opened 2016-09-28
Minor updates to odds in smirky and results output

bannanc:
This is a follow-up to pull request #149 

I fixed a small bug in sampler_smirky that was causing repeated SMIRKS patterns to be allowed. I added a check to sampler_smirky that looks to see if a new pattern empties any of the existing patterns. I also changed up a few odds to see if better sampling would happen.

There are new example runs at temperatures 0.0001 and 0.1 in each example folder:
examples/smirky/custom_Vdw
examples/smirky/custom_Bonds
examples/smirky/custom_Angles
examples/smirky/custom_Torsions

I also added a new output smarts file, which prints the smirky output to a smarts file (which could be used as input to continue sampling those patterns). The plots added in pull request #149 are also included in the new examples. 

@jchodera Here's an example plot: [custom_Bonds at 0.1](https://github.com/open-forcefield-group/smarty/blob/issue141/examples/smirky/custom_Bond/Bond_1.00e-01.pdf). They are sort of helpful in visualizing how long we have a given score.

Given the complexity in recent runs, I think we are capable of getting 100% on AlkEtOH, it is just a matter of finding the "right" odds for all the input and the "ideal" temperature for these simulations, given the examples I have now, I think 0.0001 is too low (torsions and VdW get stuck) and 0.1 is too high (creates and keeps some really weird patterns and destroys good things).

@davidlmobley and @camizanette there are a lot of changes because of all the new output files in examples, but if one of you could look at the changes in sampler_smirky to double check, that would be helpful.


davidlmobley:
@camizanette - do you have time to review the changes in `sampler_smirky` in the next couple of hours? Otherwise I'll try to do after 8 pm or in the morning.

bannanc:
I changed the name of a function and didn't change the test, for they pass now.

@davidlmobley the tests have a warning 
`Warning: Unable to parse SMARTS
Warning: [*:1]~[$([#7!-1,#8,#16])&:2]`
Do you know where the SMARTS pattern is? I think the & on the end is problematic.

davidlmobley:
Not offhand, @bannanc . And it does look odd. I'd grep for it in the tests directory; if it's not there it has to be in something under `data`.

bannanc:
> Not offhand, @bannanc . And it does look odd. I'd grep for it in the tests directory; if it's not there it has to be in something under data.

I've grepped for it in data and tests and I haven't found it yet, I'll keep an eye out and try to catch which test it is showing up in next time I run nosetests

bannanc:
I added more results, at 0.1 and 0.0001 temperatures for Vdw, Bonds, Angles, and Torsions. As of these changes I get 100% on VdW, Bonds, and Torsions. I think we're close with Angles, but may require a little more odds tweaking (hopefully with input files and not inside smirky). 

I also added fixes to filter_molecule_sets.py that @davidlmobley , @camizanette and I discussed offline.

bannanc:
@davidlmobley I got all of the changes to code here now.

I'm now rerunning the examples and tests with the updated smirky I'll add those later today or tomorrow depending how long they take to run.

smarty/sampler_smirky.py
```diff
@@ -835,7 +845,7 @@ def sample_types(self):
         env = random.choice(proposed_envList)
 
         # determine create or destroy:
-        if random.random() < 0.5:
+        if random.random() < 0.2:
```
davidlmobley, line 914: Is there a way we can determine what this ought to be? (not for this PR, just curious how to think about it...)

davidlmobley:
@bannanc - this looks good to me. I did notice randomly that one of the PDF files added in one of the early commits basically has an empty graph. 

I would push to merge this relatively soon (i.e. as soon as you get the additions in that you justmentioned) as already the change log is so extensive (mostly to informational/supporting files of course) that it's difficult to see what's changing. :)

Also, I think you may want to get back to commenting slightly more frequently. Some of the more recent code is commented a bit too sparsely in my view. Not essential to change in this PR but something to keep in mind.

Thanks for all your work on this. It's nice to see the progress!

davidlmobley:
Oops, @bannanc , now it's failing...

smarty/sampler_smirky.py
bannanc: I will continue to think about the "right" way to do this. I think in the long run there should be some way that is based on data instead of being completely randomly, "so do we need to create or remove a parameter to better fit this data?". It also seems like if there are only initial types then it should know that removing isn't allowed so we don't waste time trying things that are always not allowed?

bannanc:
> Also, I think you may want to get back to commenting slightly more frequently. Some of the more recent code is commented a bit too sparsely in my view. Not essential to change in this PR but something to keep in mind.

I have that on my list of things to look back at as I know the most recent changes have been a bit rushed as I try to get things done so I can focus on advancement. 

exploring why it's failing now...

bannanc:
> I would push to merge this relatively soon (i.e. as soon as you get the additions in that you justmentioned) as already the change log is so extensive 

The test runs take a long time to run so I will likely let them run over night and have them ready in the morning. @davidlmobley do you want to add these changes now? I'll pull from master as soon as I can see if the tests pass now.

davidlmobley:
> The test runs take a long time to run so I will likely let them run over night and have them ready in the morning. @davidlmobley do you want to add these changes now? I'll pull from master as soon as I can see if the tests pass now.

@bannanc - yes, that sounds like a good plan. Ping me when it passes and we can merge.

bannanc:
@davidlmobley 

I think this should solve the problem, there was an issue with the condition where no reference SMIRFF is provided. I added that test on purpose because I ran into a similar issue before. It says I'm up to date with master. So assuming these checks pass, this should be ready to merge. 

Note - @camizanette and I made some edits to the filter molecule script, when I have that version, I will add some more comments/doc strings and make a separate pull request with the new molecule sets.

bannanc:
@davidlmobley It's passing

davidlmobley:
@bannanc - I just updated to master since I merged a PR in the meantime. Can you ping me when it passes again?

bannanc:
@davidlmobley passed again

davidlmobley:
Thanks, @bannanc . Merging.

PR #154, opened 2016-10-01
Fix typo in score_utils

bannanc:
@camizanette This should fix the issue you were having. 
Let me know if there is still a problem. 

If your simulation is done and you just need the plots you can also call

```
from smarty.score_utils import *
create_plot_file(trajFile, plot_filename, plot_others=False)
```

where `trajFile` is your trajectory file name and `plot_filename` is where you want the plot saved. the `plot_others` plots the individual references if True. 

That should let you make plots if you have simulations that you want the plots after they're finished.


davidlmobley:
This looks good. I'll merge. For future reference, @bannanc , it's better to explain the issue which is being fixed in a way which is clear to random onlookers, not just the person who encountered the problem. :)

PR #148, opened 2016-09-25
[WIP] Add sanity checking for topological bonds

jchodera:
Discussed in #147

PR #155, opened 2016-10-03
Make fixes to impropers discussed in 147 (proper six-fold handling). 

davidlmobley:
Now handles impropers via separate registry from propers, as discussed in #147, and correctly handles the six-fold symmetrization of impropers. 

Supersedes #148 but brings in the sanity checking introduced there.


jchodera:
Thanks!

davidlmobley:
(Passes tests locally, so I'm cautiously optimistic.)

smarty/forcefield.py
```diff
@@ -1427,25 +1523,63 @@ def createForce(self, system, topology, verbose=False, **kwargs):
...
+        # Add all improper torsions to the system
+        for (atom_indices, improper) in impropers.items():
+            # Ensure atoms are actually bonded correct pattern in Topology
+            # For impropers, central atom is atom 1
+            assert topology._isBonded(atom_indices[0], atom_indices[1]), 'Atom indices %d and %d are not bonded in topology' % (atom_indices[0], atom_indices[1])
+            assert topology._isBonded(atom_indices[1], atom_indices[2]), 'Atom indices %d and %d are not bonded in topology' % (atom_indices[1], atom_indices[2])
+            assert topology._isBonded(atom_indices[1], atom_indices[3]), 'Atom indices %d and %d are not bonded in topology' % (atom_indices[1], atom_indices[3])
+
+            # Impropers are applied to all six paths around the trefoil
+            for (periodicity, phase, k) in zip(improper.periodicity, improper.phase, improper.k):
```
jchodera, line 1571: You can use [`permutations`](https://docs.python.org/2/library/itertools.html#itertools.permutations) from [`itertools`](https://docs.python.org/2/library/itertools.html) to avoid having to manually expand all permutations

davidlmobley: Ah, I figured there would be something like that but it didn't Google immediately. Nice.

smarty/forcefield.py
```diff
@@ -1462,22 +1596,40 @@ def labelForce(self, oemol, verbose=False, **kwargs):
...
+            for improper in self._impropertorsiontypes:
+                print('%64s : %8d matches' % (improper.smirks, len(getSMIRKSMatches_OEMol(oemol, improper.smirks, aromaticity_model = self.ff._aromaticity_model))))
+            print('')
 
         # Add all torsions to the output list
         force_terms = []
         for (atom_indices, torsion) in torsions.items():
             force_terms.append( ([atom_indices[0], atom_indices[1], atom_indices[2], atom_indices[3]], torsion.pid, torsion.smirks) )
+        # Add all impropers to the output list
+        for (atom_indices, improper) in impropers.items():
```
jchodera, line 1625: Same comment as above: You can use [`permutations`](https://docs.python.org/2/library/itertools.html#itertools.permutations).

jchodera:
Much better, thanks!

PR #157, opened 2016-10-05
Fixing minor bugs on filter_molecule_sets

camizanette:
After talking with @bannanc, we made some fixing in filter_molecule_sets, which include:
- changing filter_set molecules.py to filter very small molecules that we are not interested in;
- fixing a bug. 

I also incorporated the fixing made by @bannanc for the typo issue on score_utils.py and fixed one minor detail on sampler_elements regarding the statistic printing.

@davidlmobley

PR #159, opened 2016-10-14
Fixing merging problem with filter_molecule_sets.py

camizanette:
This is just a fix of an error in merging the filter_molecule_sets changes from PR #157. 
I also include my test set for smarty, which includes 4565 molecules, called DrugBank_updated*.

@davidlmobley 
@bannanc


utilities/filter_molecule_sets/filter_molecule_sets.py
```diff
@@ -66,123 +57,109 @@ def keep_molecule(mol, max_heavy_atoms = 100,
...
+                      
     # Check input files
     if (opt.input_file is None) or (opt.output_file is None):
         parser.print_help()
         parser.error("Input molecules file cannot be None")
 
     # Load and check input file
     ifs = oechem.oemolistream(opt.input_file)
-    # TODO: get input file format/flavor based on input file name?
-    flavor = oechem.OEIFlavor_Generic_Default | oechem.OEIFlavor_MOL2_Default | oechem.OEIFlavor_MOL2_Forcefield
```
davidlmobley, line 185: We don't need the forcefield flavor stuff anymore, @camizanette ? Or should this be an option to the filter tool?

utilities/filter_molecule_sets/filter_molecule_sets.py
```diff
@@ -66,123 +57,109 @@ def keep_molecule(mol, max_heavy_atoms = 100,
...
     parser.add_option('-t', '--type', type = 'string', dest = 'atomtype', default = None,
-            help = "Atom type that you do not want in your set of molecules")
+                      help = "List of atom types that you do not want in your set of molecules. Usage: gg,Se (no space between types) ")
     parser.add_option('-y', '--flavor', type = 'choice', dest = 'flavor', default = 'tripos',
-            choices = ['tripos', 'ff'],
-            help = "If you want to include different flavor - options: tripos or ff")
-
-    parser.add_option('-z', '--number', type = 'int', dest = 'number', default = 10000, help = "Limit for the number of molecules in the final set, returns the first z molecules that meet the filtering requirements")
-
-    parser.add_option('-C', '--conformers', type = 'choice', dest = 'conformers',
```
davidlmobley, line 171: No option to generate conformers anymore? Why?

davidlmobley:
Basically looks good, @camizanette (and @bannanc ) but I had a couple quick questions as to why things are done the way they are before we merge, see above. Thanks.

utilities/filter_molecule_sets/filter_molecule_sets.py
camizanette: The flavor is still there, with the option to choose between tripos or force field types.

utilities/filter_molecule_sets/filter_molecule_sets.py
camizanette: The original filter never had this option. @bannanc created this option for personal use, and when we were merging things, this option merged by mistake. If you think it's important, we can add this option.

utilities/filter_molecule_sets/filter_molecule_sets.py
```diff
@@ -195,29 +172,16 @@ def filter_molecules(input_molstream, output_molstream, allow_repeats = False,
     if not ofs.IsValid():
         parser.print_help()
```
camizanette: @davidlmobley the flavor option here.

utilities/filter_molecule_sets/filter_molecule_sets.py
davidlmobley: Ah, OK. No, I don't think it's an option we necessarily need at this stage -- it was more of a "why is this being removed" question. Thanks.

PR #160, opened 2016-10-14
Import SMIRFF force field format description from wiki draft

jchodera:
@davidlmobley drafted these nice docs describing the SMIRFF `ffxml` format in the [wiki](https://github.com/open-forcefield-group/smarty/wiki/The-SMIRFF-force-field-format), but that isn't version controlled. Ingesting them here.

cc: #133 

Preview [here](https://github.com/open-forcefield-group/smarty/blob/0a78d68d58bcee056cb731da2f49ea229037e42d/The-SMIRFF-force-field-format.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This is a simple example of how Bayesian atom type sampling using reversible-jump Markov chain Monte Carlo (RJMCMC) [1] over SMARTS types might work.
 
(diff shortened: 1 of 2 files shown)


davidlmobley:
Just got out of my (very long) call to do this and found you already did it. Thanks. Merging.

PR #162, opened 2016-10-15
Clean up docs

jchodera:
- added lots of links
- cleaned up some ambiguities
- edited code example

It looks like there are a lot more changes then there are because there are multiple sentences per line of the text file.  A future cleanup pass should separate each sentence onto its own line for better `diff` readability


davidlmobley:
Looks good, thanks.

PR #171, opened 2016-10-26
Added tests for missing bonds, angles, and torsions

### LINKED ISSUE #170: Fix checks for missing parameters
@jchodera - I was under the impression from what you wrote describing your initial code that an exception would be raised if parameters were missing for any of the major required force types in a molecule (i.e. bonds, angles, torsions, LJ), but this seems not to be the case in general. I see this being done for bonds (https://github.com/open-forcefield-group/smarty/blob/master/smarty/forcefield.py#L1234) but for none of the other force terms.

I happened across this because I was starting with a specific SMIRFF and trying to make it more general, and I found that I had certain angle terms which simply HAD no angle parameters.

We'll either need to (ASAP) implement this checking, or make our basic SMIRFFs assign generic dummy parameters to all force terms and then check that these get overridden.

jchodera:
This adds tests for missing bonds, angles, and torsions.

Fixes #170 

@davidlmobley : Can you add tests now?

### DIFF
--- a/smarty/forcefield.py
+++ b/smarty/forcefield.py
@@ -234,7 +234,7 @@ class _Topology(Topology):
 
     def angles(self):
         """
-        Get a list of all angles.
+        Get an iterator over all i-j-k angles.
         """
         if not hasattr(self, '_angles'):
             self._construct_bonded_atoms_list()
@@ -252,6 +252,10 @@ class _Topology(Topology):
         return iter(self._angles)
 
     def torsions(self):
+        """
+        Get an iterator over all i-j-k-l torsions.
+        Note that i-j-k-i torsions are excluded.
+        """
         if not hasattr(self, '_torsions'):
             self._construct_bonded_atoms_list()
 
@@ -263,6 +267,9 @@ class _Topology(Topology):
                             continue
                         for atom4 in self._bondedAtoms[atom3]:
                             if atom4 == atom2:
+                                continue
+                            # Exclude i-j-k-i
+                            if atom1 == atom4:
                                 continue
                             if atom1.index < atom4.index:
                                 self._torsions.add( (atom1, atom2, atom3, atom4) )
--- a/smarty/forcefield.py
+++ b/smarty/forcefield.py
@@ -1133,8 +1133,15 @@ def _check_for_missing_valence_terms(name, topology, assigned_terms, topological
         Atom tuples defining topological valence atomsets to which forces should be assigned
 
     """
-    assigned_set = set([ tuple( index for index in atomset ) for atomset in assigned_terms ])
-    topology_set = set([ tuple( atom.index for atom in atomset ) for atomset in topological_terms ])
+    def ordered_tuple(atoms):
+        atoms = list(atoms)
+        if atoms[0] < atoms[-1]:
+            return tuple(atoms)
+        else:
+            return tuple(reversed(atoms))
+
+    assigned_set = set([ ordered_tuple( index for index in atomset ) for atomset in assigned_terms ])
+    topology_set = set([ ordered_tuple( atom.index for atom in atomset ) for atomset in topological_terms ])
 
     def render_atoms(atomsets):
         msg = ""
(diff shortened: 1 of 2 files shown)


jchodera:
There's a bug in bond checking. Fixing now.

davidlmobley:
@jchodera - I just added tests. I added one testing function which checks to see that:
1. Missing bonded terms result in an exception
2. Missing angle terms result in an exception
3. Missing torsion terms result in an exception
4. Missing LJ (Nonbonded) parameters result in an exception

You don't seem to have added a check for item 4, so the tests currently fail unless this portion is commented out. Can you add a check for item 4?

davidlmobley:
Also, can you NOT check for missing i-j-k-i torsions in three-membered rings? https://github.com/open-forcefield-group/smarty/issues/170#issuecomment-256409575

It seems that we are not currently assigning these (which is good, as that, as you point out, makes more sense!) and neither are AMBER forcefields.

jchodera:
> You don't seem to have added a check for item 4, so the tests currently fail unless this portion is commented out. Can you add a check for item 4?

Adding now!

jchodera:
> You don't seem to have added a check for item 4, so the tests currently fail unless this portion is commented out. Can you add a check for item 4?

Added a check for Lennard-Jones terms.

What should we do about charges? I think we can just ensure that every molecule in the `Topology` has a corresponding `OEMol` provided. Does that make sense?

jchodera:
> Also, can you NOT check for missing i-j-k-i torsions in three-membered rings? 

`i-j-k-i` torsions are now excluded.

jchodera:
Travis is failing to get the openeye toolkits:

```
3.23s$ pip install --pre -i https://pypi.anaconda.org/openeye/label/beta/simple openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
Collecting openeye-toolkits
  Downloading https://pypi.anaconda.org/openeye/label/beta/simple/openeye-toolkits/2016.10b7/OpenEye-toolkits-2016.10b7.tar.gz
Collecting OpenEye-toolkits-python2.7-ubuntu-12.04-x64==2016.10b7 (from openeye-toolkits)
  Could not find a version that satisfies the requirement OpenEye-toolkits-python2.7-ubuntu-12.04-x64==2016.10b7 (from openeye-toolkits) (from versions: 2016.10b1, 2016.10b2, 2016.10b3, 2016.10b4, 2016.10b5, 2016.10b6)
No matching distribution found for OpenEye-toolkits-python2.7-ubuntu-12.04-x64==2016.10b7 (from openeye-toolkits)
```

davidlmobley:
> What should we do about charges? I think we can just ensure that every molecule in the Topology has a corresponding OEMol provided. Does that make sense?

Sounds good to me.

On travis -- sometimes OpenEye's beta builds go up not simultaneously, so probably the appropriate beta is not up yet. It'll probably be up in a day or two.

davidlmobley:
@jchodera - is this at the point where it should be merged if tests pass locally?

jchodera:
Yes! It works for me, except for an absolutely insane number of OpenEye warnings:

```
Warning: data type mismatch found when using generic data
```

jchodera:
But please test locally on your machine! I've only tested `smarty.tests.test_forcefield`.

davidlmobley:
Yes, I've gotten those warnings too; I don't know where they come from (they arrived a while back when you added a larger number of tests, I think). I'll test.

jchodera:
It has to do with replacing tag data with a different type:
https://docs.eyesopen.com/toolkits/python/oechemtk/genericdata.html

I wonder if it's connected to the fractional bonds.

davidlmobley:
> I wonder if it's connected to the fractional bonds.

Hmm, it could be. I do use `SetData`, but I just do it to store back into bond orders the same data that was originally there (without changing its type); it's not clear to me why that would trigger this warning. (This is to allow charge calculation to NOT modify the original molecule.)

So, this is likely ignorable. I wish it were easier to suppress specific OE warnings though.

davidlmobley:
Tests work for me though (I ran all of them!). I'll merge.

PR #172, opened 2016-10-29
Update AlkEthOH example; add code testing creation of OEMols from elements, connectivity, bond orders

davidlmobley:
As requested by @jchodera , updates the AlkEthOH example and README.md to show a successful test of 500 iterations at T=0.

Also, unrelatedly, adds an IPython notebook under `examples/testing` which checks whether OpenEye molecules can be successfully and unambiguously created (even for charged molecules) from elements, atom names, connectivity, and integer bond orders alone. The verdict seems to be yes. This relates to discussion in https://github.com/pandegroup/openmm/issues/1443#issuecomment-256785686. The relevant notebook is OE_bondorder_testing.ipynb

PR #173, opened 2016-10-29
Fix py3 issue with exception handling

jchodera:


jchodera:
I _think_ this fixed the issue noted in #172

davidlmobley:
Seems to be resolved, thanks!

Issue #179, opened 2016-11-03
Replace HarmonicAngleForce with AngleForce (and likewise for bonds?) in SMIRFF


davidlmobley:
We need to replace the `HarmonicAngleForce` SMIRFF tag with `AngleForce`, which would default to harmonic unless otherwise specified ( to allow the format to support other forms if desired), as per @jchodera discussion with Swope and others at IBM.

@jchodera - presumably we should do the same for bonds as well, to potentially allow support for Morse potentials, etc.?

jchodera:
Yes, I think we should generalize bonds and angles this way.

We should also generalize `NonbondedForce` to permit other forms (e.g. Buckingham exp-6, Halgren buffered 14-7, etc.).

davidlmobley:
> We should also generalize NonbondedForce to permit other forms (e.g. Buckingham exp-6, Halgren buffered 14-7, etc.).

Yes. I didn't mention that here, as `HarmonicAngleForce` and `HarmonicBondForce` are tags in the format which need changing, whereas the `NonbondedForce` tag doesn't need changing (I don't think) we just need to change how it's implemented. In other words, changing bonds/angles requires modifying our current code whereas handling other nonbonded forms just requires generalizing/extending. So I figured that was a separate issue.

(I'm working up a PR to the format spec which would address these issues.)

jchodera:
Could address #42 (versioning) as well!

davidlmobley:
On versioning - I commented on #42.

jchodera:
Addressed in https://github.com/openforcefield/openforcefield/pull/86

Issue #175, opened 2016-11-01
Decouple smarty from ForceField/SMIRFF for versioning


davidlmobley:
We'll need to move to get the Smirks force field format into a separate repo for versioning reasons -- specifically, we'll need to start versioning the specifications for the format, and having the parsing code check that the version of the XML being parsed is consistent with the version of the parser doing the parsing. But @jchodera points out that we don't want to have to cut a new version of the spec every time we change something in smarty or smirky, hence a need to separate the two now.

@jchodera says he will tackle this soon.

davidlmobley:
@jchodera - now that #239 is merged you should be able to proceed with this.

jchodera:
Unless there are objections, the new repository will be called https://github.com/open-forcefield-group/smirnoff

davidlmobley:
Excellent, @jchodera !

I'm mostly offline today but will try and scroll through my mass of github notifications really quickly to see if there is anything urgent.

jchodera:
I'm trying to see if there's a way to preserve commit history during the migration. Stay tuned!

bannanc:
@jchodera @davidlmobley I think this can be closed since smarty/smirnoff99Frosst/openforcefield should all have their own versions now correct?

davidlmobley:
Yes. Thanks.

PR #180, opened 2016-11-03
Adjust SMIRFF format spec to address generalization of format

davidlmobley:
- [x] Discuss how other mixing/combination rules would be handled
- [x] Discuss how other functional forms for the nonbonded term would be selected
- [x] Discuss how other functional forms for valence terms such as angle terms would be selected (and mention switch to `AngleForce` rather than `HarmonicAngleForce`, as per #179 
- [x] Remove id/parent_id from formal spec, just mentioning `id` in a usage example
- [x] Put in mention of GB parameters
- [x] Add line breaks for better change logs

@jchodera , anything else I'm missing?


The-SMIRFF-force-field-format.md
```diff
@@ -1,33 +1,40 @@
...
+The SMIRFF format provides XML `ffxml` files that are parseable by the `ForceField` class of the `smarty.forcefield` module. 
+These encode parameters for a force field based on a SMIRKS-based specification of the chemical environment the parameters are to be applied to. 
+The file has tags corresponding to OpenMM force terms (`HarmonicBondForce`, `HarmonicAngleForce`, `PeriodicTorsionForce`, etc., as discussed in more detail below); these specify units used for the different constants provided for individual force terms, for example (see the [AlkEthOH example ffxml](https://github.com/open-forcefield-group/smarty/blob/master/smarty/data/forcefield/Frosst_AlkEtOH.ffxml)):
 ```XML
    <HarmonicAngleForce angle_unit="degrees" k_unit="kilocalories_per_mole/radian**2">
 ```       
 which introduces following `Angle` terms which will use units of degrees for the angle and kilocalories per mole per square radian for the force constant.
 
 Under each of these force terms, there are tags for individual parameter lines such as these:
 ```XML
```
jchodera, line 17: I'd add the `<AngleForce>...</AngleForce>` tags here to make it very clear that the `<Angle.../>` tags have to go inside this block.

The-SMIRFF-force-field-format.md
```diff
@@ -53,64 +60,74 @@ For compatibility, the size property of an atom can be specified either via prov
...
 ```
 
-**AMBER functional forms drop the factor of 2 in the angle energy term, which we elect not to do here.** Thus, comparing a SMIRFF file to a corresponding AMBER parameter file or .frcmod will make it appear that force constants here are twice as large as they ought to be.
+**AMBER functional forms drop the factor of 2 in the angle energy term, which we elect not to do here.**
+Thus, comparing a SMIRFF file to a corresponding AMBER parameter file or .frcmod will make it appear that force constants here are twice as large as they ought to be.
 
 ### PROPER TORSIONS
 
 Torsions are implemented as a [`PeriodicTorsionForce`](http://docs.openmm.org/7.0.0/userguide/theory.html#periodictorsionforce) tag with child tags for `Proper` (discussed here) and `Improper` (discussed below) parameters, for example:
 ```
```
jchodera: Forgot `XML` after '```'

jchodera:
Are you planning to change the `<Harmonic*Force.../>` tags to drop the `Harmonic` part in this PR?

davidlmobley:
@jchodera :

> Are you planning to change the `<Harmonic*Force.../>` tags to drop the Harmonic part in this PR?

I was going to wait til we change the code to do that. Right now it's listed here under things we are GOING to do. When we do it, then I'll drop the `Harmonic`.

The-SMIRFF-force-field-format.md
davidlmobley: Thanks.

jchodera:
How about this plan:
- Commit this PR once ready
- Create a new branch for the new updates (e.g. `0.2.0`) and make the updates to the spec
- I'll then update the code to match the spec
- We then merge the whole branch into `master`.

The-SMIRFF-force-field-format.md
davidlmobley: Wait, hold on -- that's different from what I thought we were saying. I thought we were saying we'd just be replacing the `HarmonicAngleForce` tag with an `AngleForce` tag and whether it is harmonic or not would be specified by an attribute of the `SMIRFF` tag, as per what I wrote in the new docs:

> ... Selection of the type of angle force applied would be handled in a similar manner, via an `AngleForce="harmonic"` tag or similar. [This feature is being planned.](https://github.com/open-forcefield-group/smarty/issues/179)

jchodera:
Actually, we may want to split off the SMIRFF forcefield stuff to a new repo before doing that.

davidlmobley:
@jchodera - regarding splitting into separate repo, yes, that was what we discussed by phone the other day. I made an issue and assigned it to you, but let me know if you want me to do it instead. 
https://github.com/open-forcefield-group/smarty/issues/175

(Whoever does it needs to get travis going, etc., on the new repo, as well as connect it up to omnia).

jchodera:
Yep, I've got it slated for the weekend. Do we need it earlier than that?

davidlmobley:
No, weekend is fine. But I'd be happy to do it myself if it would free up time for you to do the property calculator for densities or GB solvation.

The-SMIRFF-force-field-format.md
davidlmobley: @jchodera - any further comments on this?

jchodera: > Wait, hold on -- that's different from what I thought we were saying. I thought we were saying we'd just be replacing the HarmonicAngleForce tag with an AngleForce tag and whether it is harmonic or not would be specified by an attribute of the SMIRFF tag, as per what I wrote in the new docs:

We're in agreement here!  But it still says `HarmonicAngleForce` and not `AngleForce`. Same thing with `HarmonicBondForce`.

jchodera:
I've fallen behind again here, but I realized I needed to [finish the GBSA PR](https://github.com/open-forcefield-group/smarty/pull/182) before peeling of the SMIRFF code to another repo.

jchodera:
You should be good to rebase or merge from master and then we can merge this in.

The-SMIRFF-force-field-format.md
davidlmobley: Yes, totally agree. Just need to make the associated change in the code and then we can do that. I'll rebase/merge.

PR #182, opened 2016-11-11
Additions for GBSA support

jchodera:
Checking in my progress on GB support from a while ago. Still need to finish.

The syntax looks like:
```XML
 <GBSAForce model="OBC1" radius_units="nanometers" sasa_penalty="5" sasa_penalty_units="cal/mol/angstroms**2">
   <Atom smirks="[#1:1]" radius="0.12" scale="0.85"/>
   <Atom smirks="[#1:1]~[#6]" radius="0.13" scale="0.85"/>
   <Atom smirks="[#1:1]~[#8]" radius="0.08" scale="0.85"/>
   <Atom smirks="[#1:1]~[#16]" radius="0.08" scale="0.85"/>
   <Atom smirks="[#6:1]" radius="0.22" scale="0.72"/>
   <Atom smirks="[#7:1]" radius="0.155" scale="0.79"/>
   <Atom smirks="[#8:1]" radius="0.15" scale="0.85"/>
   <Atom smirks="[#9:1]" radius="0.15" scale="0.88"/>
   <Atom smirks="[#14:1]" radius="0.21" scale="0.8"/>
   <Atom smirks="[#15:1]" radius="0.185" scale="0.86"/>
   <Atom smirks="[#16:1]" radius="0.18" scale="0.96"/>
   <Atom smirks="[#17:1]" radius="0.17" scale="0.8"/>
 </GBSAForce>
```


jchodera:
The `GBSAForceGenerator` code hasn't been altered from `NonbondedForce` yet. Working on that now.

davidlmobley:
Spec changes look good. LMK when you are ready for code review, @jchodera .

jchodera:
Will do! Still coding for now...

jchodera:
I've been going through the changes required for this addition. It's going to be a bit more involved than expected because we'll need to distribute a version of the `CustomGBForce` implementations of the GBSA models we want to support until we can get some additional changes merged into OpenMM since an overhaul of the GBSA forces for `simtk.openmm.app` ended up making them a lot less functional for those (like us) intending to modify the parameters fed to them away from AMBER parameters. I will have to either undo some of these changes or backport some bugfixes to an earlier version of the GBSA forces.

@davidlmobley : What is the most important immediate use case? Would it be just OBC2?

davidlmobley:
@jchodera :

> What is the most important immediate use case? Would it be just OBC2?

Yes! Really we just need to get the basic framework for support of SOME GB model (OBC2 is a good choice) in place. Then, we can later deal under the hood with being able to support more GBSA models.

jchodera:
I've managed to get HCT, OBC1, and OBC2 in, but there is one more major issue to fix with how (radius, screen) are translated into what parameters used to be `sr` and `or` in the `gbsaforces.py`. I hadn't realized the energy expressions were encoded in terms of a shifted radius and a scaled radius rather than the original radius and scale parameters---this will take a few min to figure out, and are required for numerical correctness, but I want to at least make sure the code is doing the right things as far as parameter assignment right now.

In running the tests, I've been getting these errors:
```
ERROR: Test setup of a molecule which activates partial bond order code.
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/tests/test_forcefield.py", line 448, in test_partial_bondorder
    system = ff.createSystem(topology, [mol], chargeMethod = 'OECharges_AM1BCCSym', verbose = verbose)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/forcefield.py", line 947, in createSystem
    self._assignPartialCharges(molecule, oechargemethod)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/forcefield.py", line 856, in _assignPartialCharges
    partial_bondorders.append( bond.GetData("WibergBondOrder"))
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/openeye/oechem.py", line 7473, in GetData
    return _oechem.OEBase_GetData(self, *args)
ValueError: GetData: WibergBondOrder not found.
```
Any idea what is going on here?

davidlmobley:
@jchodera - are you running OpenEye toolkits 2016.10 or later? The Wiberg bond orders were in the October toolkits versions (and thereafter I believe). 

(This is used in tests -- we have a test that looks to see whether partial (Wiberg) bond orders are working correctly by looking at benzene or something similar, using an aromatic C~C bond length which is interpolated between single and double. If the OE AM1-BCC calculation doesn't assign partial bond orders (which it did beginning in the October release) then the test fails.)

(I suspect this is an issue with your local install since these tests pass on my machine and on Travis-CI, at least up til a week or so ago when they started timing out.)

davidlmobley:
If the above is NOT an issue with your OpenEye toolkits, raise a separate issue and I can troubleshoot.

jchodera:
> @jchodera - are you running OpenEye toolkits 2016.10 or later? 

That was the issue! Thanks!

Now I'm seeing:
```
ERROR: Test evaluation of energies from parm@frosst ffxml files versus energies of equivalent systems.
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/tests/test_forcefield.py", line 358, in test_smirff_energies_vs_parmatfrosst
    results = compare_molecule_energies( prmtop, crd, forcefield, mol, verbose = verbose )
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/forcefield_utils.py", line 270, in compare_molecule_energies
    smirfftop, ambersys, smirffsys, amberpos, verbose = verbose, skip_assert = skip_assert, skip_improper = skip_improper )
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty.open-forcefield-group/smarty/forcefield_utils.py", line 208, in compare_system_energies
    syscheck.check_force_parameters(skipImpropers = skip_improper)
TypeError: check_force_parameters() got an unexpected keyword argument 'skipImpropers'
```

jchodera:
Also, about a billion warnings like
```
Warning: data type mismatch found when using generic data
```

davidlmobley:
Update your `openmoltools`.

Maybe I did not remember to update these in the package requirements?

davidlmobley:
So, warnings like this:
> Warning: data type mismatch found when using generic data

Those became normal when I think we started accessing the Wiberg bond orders, I think. They're not functionally important, but I need to figure out how to disable them. I need to create an issue but am on the way out the door; you can create one and assign it to me if you like.

jchodera:
Ah, that was my fault. I neglected to update `openmoltools` to 0.7.3.

jchodera:
Remaining to be done:
- [ ] Fix how forcefield parameters (`radius` and `scale`) are translated into parameters in the `CustomGBForce` energy expressions. This requires either retooling of the custom energy expressions in `gbsaforces.py` or the way parameters are set (which would be harder to accommodate without making the `forcefield.py` code significantly more complicated). 
- [ ] Expose the ability to control the SASA penalty
- [ ] Generalize the SA models to allow us to extend this to other SA models in the future (which may have different numbers of parameters) like EEF1
- [ ] Add GBn and GBn2 models
- [ ] Add additional tests of variants of specifying or omitting some GBSA parameters to make sure we trigger informative exceptions when we omit some parameters

At this point, it would be useful to get feedback on the [current `<GBSAForce/>` tag spec](https://github.com/open-forcefield-group/smarty/blob/8bb8c2ab510770088d7a3db45327cf76525f6f45/The-SMIRFF-force-field-format.md#gbsa-parameters) to make sure it exposes what is needed.

davidlmobley (commented):
I'm on a phone at the moment so can't review new spec in detail right now but can try to do so after 8   Pacific, or if you need feedback sooner, @jchodera you can tag Caitlin and see if she is available to look at the spec.

jchodera:
I won't be able to tackle the rest until the morning, so no rush!

davidlmobley:
@jchodera - the GBSA spec looks good to me except a couple of minor things:

Relating to alpha, beta, gamma:
> Units are for these per-particle parameters (such as radius_units) optionally specified in the <GBSAForce/> tag.

It's not actually clear to me from this what allowable units are for alpha, beta, and gamma. The link is not that helpful since it just leads to a reference. Can we at least say "has units of length" (or whatever it should have.)? 

Also (and maybe this is just my lack of knowledge):
> The sa_model attribute specifies the solvent-accessible surface area model ("SA" part of GBSA), which can currently only be ACE or omitted (in which case the SA term is omitted)

What is "ACE"? Spell out on first use? 

Otherwise, the spec looks great to me!

jchodera:
> It's not actually clear to me from this what allowable units are for alpha, beta, and gamma. The link is not that helpful since it just leads to a reference. Can we at least say "has units of length" (or whatever it should have.)?

I'm not sure what the units of `alpha`, `beta`, and `gamma` for GB(neck) models are either. We'd have to look in the original references to see if these parameters have units. If they do have units, they'd be interpreted in the OpenMM unit system (e.g. nanometers) unless `alpha_unit` was specified. I'll do some digging tomorrow.

> What is "ACE"? Spell out on first use?

I believe the "ACE" model is [this model from Schaefer and Karplus](http://pubs.acs.org/doi/abs/10.1021/jp9521621), for "analytical continuum electrostatics".

davidlmobley:
> I'm not sure what the units of alpha, beta, and gamma for GB(neck) models are either. We'd have to look in the original references to see if these parameters have units. If they do have units, they'd be interpreted in the OpenMM unit system (e.g. nanometers) unless alpha_unit was specified. I'll do some digging tomorrow.

Haha, OK. Are you sure you don't want to just move this into the category of "we will eventually support..." and punt it for now? The main point at this point, at least as I understood it, was to at least get basic GBSA support for one or more models in so that we could flesh out the spec; you seem to have achieved that.

Thanks for the pointer on ACE.

jchodera:
Quick update: Nearly finished with this, but finishing some additional cleanup/refactoring to make the spec more extensible:
* Having separate attributes for the surface area penalty and the units for the surface area penalty didn't make sense---it makes much more sense to just let units be specified in the surface area penalty attribute, like `5*calories/angstrom**2`.
* I've refactored things to be a bit more general in that you can specify a `gb_model` and an `sa_model` for the GB and SA parts, respectively, and each model may have a few global parameters (and per-particle parameter unit specifiers) that can be specified separately.

Will get this wrapped up today.

ctk3b:
> I'm not sure what the units of alpha, beta, and gamma for GB(neck) models are either. We'd have to look in the original references to see if these parameters have units.

In case this wasn't resolved yet, I just went back to take a peek at the [reference](http://pubs.acs.org/doi/abs/10.1021/ct600085e). Alpha/beta/gamma are unitless scaling terms (Table 1) just as they are in the OBC models.

jchodera:
I think the draft implementation is now complete. Please take a look at the updated [GBSA spec](https://github.com/open-forcefield-group/smarty/blob/cdf831b42f929617b968293d2ffd086494a5b2c3/The-SMIRFF-force-field-format.md#gbsa-parameters) to make sure you're happy with it.

jchodera:
Woo! Tests pass.

Note that I haven't tested GBSA energies for correctness (that's hard)---I've just made sure the format can be parsed and the energies aren't crazy.

davidlmobley:
@jchodera - this looks excellent to me, and I don't see any problems with the spec. OK to merge? 

Then we could proceed to split off `smarty.forcefield` in a separate PR. You still want to be the one to do this, right?

davidlmobley:
Actually, I see it's OK to merge since you changed the title. I'll go ahead. :)

jchodera:
> Then we could proceed to split off smarty.forcefield in a separate PR. You still want to be the one to do this, right?

Yep! On it.

davidlmobley:
@jchodera - one other thought on that. There are various things in `utilities`, `examples`, etc. that will belong in one repo or the other. I'm happy to handle moving those to where they ought to be, to save you time, after you get the basic skeleton up. That will also provide me with the impetus to do some cleanup/re-check documentation to make sure everything is well organized and clearly documented (I think it is, but it never hurts to re-check). I wouldn't want you to get bogged down in the reorganization because you couldn't tell where something belonged...

Thanks.

Issue #184, opened 2016-11-21
Make a utility taking systems into AMBER/GROMACS


davidlmobley:
Make a utility or example which will take SMIRFF system and go to AMBER `prmtop` and `crd` and/or GROMACS `top` and `gro`, as per https://github.com/choderalab/yank/issues/544#issuecomment-262032469

I can rely on github.com/mobleylab/solvationtoolkit for system prep, so really this means just taking a system which is already prepped (`.pdb` and `.mol2` files for the system and components available), parameterizing with SMIRFF, and going into GROMACS or AMBER format. Since the conversion from OpenMM to AMBER/GROMACS can be handled by `ParmEd`, I think this should be fairly trivial.

jchodera:
Do we want to use TIP3P water or parm@frosst water? And would we want to use Amber or SMIRFF formats for the water?

You'll want to include as input:
* PDB file for the system to read in Topology and positions
* Mol2 files with partial charges for small molecule and solcent (if parameterizing solvent box)

I would save the serialized OpenMM System XML as well, just in case that is useful.

davidlmobley:
@jchodera - how would I save the serialized OpenMM System XML? Am not familiar with that.

Without some new development, I think we can only do AlkEthOH flexible water if we take the whole system from SMIRFF. We don't have constraints yet, nor do we have TIP3P in SMIRFF.

Potentially I could figure out how to do a SMIRFF solute in conventional TIP3P if I bring in TIP3P from AMBER (which would also allow for constraints), though this will take me longer. As a first pass implementation I plan to just provide a way to take an arbitrary all-SMIRFF system into AMBER and GROMACS.

jchodera:
Actually, I wonder if we can easily graft in a reader for OpenMM XML and PDB (much like the amber or Gromacs readers)  so you can skip the ParmEd step. Let me ask @andrrizzi in the morning.

jchodera:
> @jchodera - how would I save the serialized OpenMM System XML? Am not familiar with that.

```python
# serialize system to a string
from simtk.openmm import XmlSerializer
serialized_system = XmlSerializer.serialize(system)
# now write that string to a file
```

> Without some new development, I think we can only do AlkEthOH flexible water if we take the whole system from SMIRFF. We don't have constraints yet, nor do we have TIP3P in SMIRFF.

We can probably easily add constraints. Did that make it into the spec, or
is there just an issue?

[...]

PR #187, opened 2016-11-22
Add conversion to AMBER and GROMACS formats

### LINKED ISSUE #186: Multi-component systems lose charges in subsequent components
I'm working on #184, prototyping conversion of the cyclohexane/ethanol mixed system which is provided here into AMBER via ParmEd, and uncovered a bug where subsequent molecules (after the first one) in a system end up with zero charges. I'll troubleshoot.

davidlmobley:
Here I'll add conversion to AMBER and GROMACS formats as per #184 :
- [x] Convert to AMBER format (`forcefield_utils.save_system_to_amber`)
- [x] Convert to GROMACS format (`forcefield_utils.save_system_to_gromacs`)
- [x] Add ParmEd to requirements
- [x] Add test validating energies before and after conversion

Several other issues were fixed in the process:
- [x] Extraneous print statement removed
- [x] Better debug info for a nonbonded force exception
- [x] Fix #186 bug with charges
- [x] Fix documentation issue noted in #186 

Other things to fix as well:
- [x] Add a test that multiple instances of a particular molecule in a system get correct charges (not just the first molecule) per #186


davidlmobley:
I'm continuing to work on conversion, but there is some sort of issue with exclusions appearing twice (?) in the resulting AMBER prmtop and crd files, so I've raised an issue on parmed: https://github.com/ParmEd/ParmEd/issues/784

davidlmobley:
Currently I'm running into an issue where one of the exclusions in ethanol has a different charge-charge interaction parameter when round-tripping from ParmEd versus originally in our OpenMM system. 
```
AssertionError: Error: NonBondedForce Exception atom (2702, 2705) has charge values of -0.026858 and -0.022381 elementary charge, respectively.
```

This is a OH-HC 1-4 interaction. The product of the charges on the atoms in both topologies gives me the -0.0268 value, so I'm not sure where the -0.022 value is coming from. I wonder if this is somehow related to the alternate value of the Coulomb constant used in AMBER. Checking.

davidlmobley:
Differences in Coulomb constant don't seem to explain it; the difference is far too small. 332.0522 in AMBER, and the most different other code has 332.0637 or so. 

Also, it's odd that this only occurs here and in none of our prior tests. I wonder if I have some sort of lingering issue with copying charges, where the forces in OpenMM are somehow being applied with slightly different charges than are originally in the molecules or some such. Investigating...

davidlmobley:
Aha! The difference is 0.833333, which is the 1-4 electrostatics scale factor. So ParmEd or OpenMM is picking up an extra factor of 0.83333 in the conversion, it seems. Investigating.

davidlmobley:
@swails, @jchodera -- I'm creating an OpenMM system from our SMIRFF format via code by the below, writing out to AMBER parameter/crd, reading back in, and cross-checking the energy from the AMBER files vs the original system to make sure it is still the same, and I'm picking up or missing a factor of 0.83333 in the 1-4 Coulomb interactions, probably inside ParmEd somewhere though it could be in OpenMM (it's not in this package since we don't work with the 1-4 interactions other than just setting the scale factor, and we DO normally reproduce AMBER energies -- so it seems to be the ParmEd round trip which is messing things up). Any insights as to where this could be?

Here's how I create the system and compare energies:

```python
from smarty.forcefield import *
import parmed
import openeye.oechem as oechem
from simtk.openmm.app import PDBFile
pdbfile = PDBFile('cyclohexane_ethanol_0.4_0.6.pdb')
ff = ForceField('../../smarty/data/forcefield/Frosst_AlkEtOH.ffxml')

mols = []
flavor = oechem.OEIFlavor_Generic_Default | oechem.OEIFlavor_MOL2_Default | oechem.OEIFlavor_MOL2_Forcefield
for molname in ['cyclohexane', 'ethanol']:
    ifs = oechem.oemolistream(molname+'.mol2')
    ifs.SetFlavor( oechem.OEFormat_MOL2, flavor)
    mol = oechem.OEGraphMol()
    while oechem.OEReadMolecule(ifs, mol):
        oechem.OETriposAtomNames(mol)
        mols.append(oechem.OEGraphMol(mol))

system = ff.createSystem( pdbfile.topology, mols)
#Swap in positions
integrator = openmm.VerletIntegrator(1.0 * unit.femtoseconds)
context = openmm.Context(system, integrator)
context.setPositions(pdbfile.positions)
# Create parmed Structure
structure = parmed.openmm.topsystem.load_topology( pdbfile.topology, system, pdbfile.positions)
# Write AMBER parameter/crd
structure.save('system.prmtop', overwrite=True)
structure.save('system.crd', format='rst7', overwrite=True)

# Evaluate energy of OpenMM system and AMBER system to check that they are still the same
from smarty.forcefield_utils import *
parm = parmed.load_file('system.prmtop', 'system.crd')
ambersys = parm.createSystem(nonbondedMethod= app.NoCutoff, constraints = None, implicitSolvent = None)
groups0, groups1, energy0, energy1 = compare_system_energies( parm.topology, pdbfile.topology, ambersys, system, pdbfile.positions)
```

The energy comparison at the end results in an error for the OH~HC 1-4 interaction, where the scaled product of the charges output by OpenMM's `getExceptionParameters` differs by a factor of 0.83333 depending on whether the system is coming from ParmEd output to AMBER format read back in by ParmEd, or from the original system without any output.

davidlmobley:
@swails and @jchodera - I serialized the XML for the relevant OpenMM systems and stored them to disk; I'm attaching them here (they are also in the most recent commit). @swails , this should allow you to grab the OpenMM systems in question without having to have the OpenEye toolkits.

If you get the latest openmoltools from GitHub (github.com/choderalab/openmoltools), you should be able to load these files and then use:
```python
from simtk.openmm.app import PDBFile
from simtk.openmm import XmlSerializer
from simtk.openmm.app import Topology
from simtk import openmm
# I might be missing a couple of imports here
smirff_system = XmlSerializer.deserialize(serialized_system_smirff)
amber_system = XmlSerializer.deserialize(serialized_system_amber)
pdbfile = PDBFile('cyclohexane_ethanol_0.4_0.6.pdb')

# Do an energy comparison
timestep = 1.0 * unit.femtoseconds
integrator0 = simtk.openmm.VerletIntegrator( timestep )
integrator1 = simtk.openmm.VerletIntegrator( timestep )
platform = simtk.openmm.Platform.getPlatformByName("Reference")
positions = pdbfile.positions
simulation0 = app.Simulation( pdbfile.topology, smirff_system, integrator0, platform = platform )
simulation0.context.setPositions(positions)
simulation1 = app.Simulation( parm.topology, amber_system, integrator1, platform = platform )
simulation1.context.setPositions(positions)
syscheck = system_checker.SystemChecker( simulation0, simulation1 )
groups0, groups1 = syscheck.check_energy_groups(skip_assert = False)
energy0, energy1 = syscheck.check_energies(skip_assert = False)
```
to cross-compare energies. 

At this point, this does NOT yet reproduce my earlier bug specifically; instead, I actually get an earlier error which relates to a large difference in nonbonded energy difference (though possibly it's from the same problem). So at this point, I tentatively think there's a separate bug in serialization of the XML.

I have not yet cross-compared the XML files.

[XML_systems.zip](https://github.com/open-forcefield-group/smarty/files/607967/XML_systems.zip)

swails:
@davidlmobley -- can you make a file that I can read back in to create a `Topology` in ParmEd?

If you go back to your original file, after you use `parmed.load_topology`, save the resulting `Structure` to a mol2 file and send that to me.  mol2 files contain all of the information necessary to rebuild the Topology faithfully.  Then I can use that to create a `Topology`, deserialize the `System`, and try to reproduce the problem.

jchodera:
A PDB file will also work for OpenMM `Topology` objects.

swails:
> and I'm picking up or missing a factor of 0.83333 in the 1-4 Coulomb interactions, probably inside ParmEd somewhere though it could be in OpenMM

It's probably in ParmEd.  If you get me the mol2 file and `System` XML file, I should hopefully be able to nail down what's happening.

swails:
> A PDB file will also work for OpenMM Topology objects.

Mol2 files are better for this test -- they store bond info (with PDBs, ParmEd would have to guess at it).

davidlmobley:
@swails - sorry, I meant to attach the PDB but forgot (that's where I'm getting the Topology from). 
[cyclohexane_ethanol_0.4_0.6.pdb.zip](https://github.com/open-forcefield-group/smarty/files/608121/cyclohexane_ethanol_0.4_0.6.pdb.zip)

Note everything IS on GitHub -- see https://github.com/open-forcefield-group/smarty/tree/issue184/examples/testing/test_system_conversion where you can pull the IPython notebook I'm using for this and all the supporting files. The code I pasted above is at the bottom of the notebook.

davidlmobley:
@swails - I'm also happy to save a .mol2 of the topology and will attempt to do so now.

davidlmobley:
@swails - here's a .mol2 of the `Structure` created via 
```python
structure = parmed.openmm.topsystem.load_topology( pdbfile.topology, system, pdbfile.positions)
structure.save('swails.mol2')
```
which hopefully is what you're after.
[swails.mol2.gz](https://github.com/open-forcefield-group/smarty/files/608129/swails.mol2.gz)

davidlmobley:
@swails - just wanted to make sure you saw the above, though I understand if you're out for the holiday weekend.

swails:
I've seen it. I'll try to get to it tomorrow or Sunday. 

Happy holidays!

--
Jason M. Swails

davidlmobley:
@swails - any success?

swails:
Hi @davidlmobley -- sorry it took me awhile to get to this.  I did finally manage to find some time, and I have a fix for the duplicate exception parameter error in ParmEd/ParmEd#789.  Give that code a try when you get a chance.  Some *really* old code that I had completely forgotten existed was causing the problem.

I doubt that it fixed the 1/1.2-factor error in the 1-4 electrostatic calculation, but that will hopefully not be as hard to find.

davidlmobley:
Thanks, @swails - I probably can't try today but will likely try asap tomorrow, as I need this to proceed with something I need for a paper. :)

davidlmobley:
Note to self: I can no longer reproduce my 1-4 scaling factor issue; in fact, currently, a round-trip of the system to AMBER format and back results in the same energy, as does serialization/deserialization from OpenMM. So, everything seems good now.

Proceeding.

davidlmobley:
Currently running into issues after converting to GROMACS format; I seem to have found a ParmEd bug (https://github.com/ParmEd/ParmEd/issues/791 ). Otherwise I am close to done with this.

davidlmobley:
ParmEd https://github.com/ParmEd/ParmEd/pull/794 fixes these issues, so now I can reproduce energies before/after saving to AMBER and GROMACS format. I'll clean up my notebook that does this, commit it, then get tests, etc. included.

swails:
woohoo!

davidlmobley:
Tests will likely fail because of needing a new `parmed` release incorporating recent bugfixes.

davidlmobley:
Yes, currently I have all the issues fixed except the problem with the older version of `parmed`, so this is waiting on a new release. 

However, I'd also like to have a test comparing energies against AMBER on a complete system (which would catch bugs in handling of additional molecules after the first, as in #186).  I'm trying to figure out if I can get a parm99 parameterized system without too much trouble for cross-comparison (at this point all the full systems I have use GAFF).

davidlmobley:
@limn1 or @bannanc - could one of you do code review on this really quickly sometime today to see if you see any issues/problems? Especially on `forcefield.py` and `forcefield_utils.py`; there are a variety of other files modified but these shouldn't be important to review.

davidlmobley:
(Tests won't pass until after https://github.com/omnia-md/conda-recipes/pull/655 is merged, presumably. Though, they pass locally since I have that version of ParmEd installed.)

bannanc:
@davidlmobley I can code review, I'm heading to a seminar soon so it might be later this afternoon.

smarty/forcefield.py
```diff
@@ -1858,11 +1857,23 @@ def createForce(self, system, topology, nonbondedMethod=NoCutoff, nonbondedCutof
...
+                    # Retrieve NB params for reference atom (charge not set yet)
                     [charge, sigma, epsilon] = force.getParticleParameters(atom_index)
-                    force.setParticleParameters(atom_index, atom.GetPartialCharge(), sigma, epsilon)
+                    # Look up the charge on the atom in the reference molecule
+                    charge = charge_by_atom[atom_index]*unit.elementary_charge
 
+                    # Set parameters for equivalent atom in other instance of
+                    # this molecule
+                    force.setParticleParameters(map_atom_index, charge, sigma, epsilon)
         # TODO: Should we check that there are no missing charges?
```
bannanc, line 1877: You still have a TODO, I don't know if this is something you wanted to check in this pull request?

bannanc (approved):
@davidlmobley  
All the changes look good. 
I made a comment on one of the TODOs, but it looks like there are still quite a few TODOs in forcefield.py. 
Assuming you're intentionally saving those for future pull requests then this is good to go.

smarty/forcefield.py
davidlmobley: Thanks. That's been there for a while, from John's initial draft of this code. At some point I may add a check that every molecule has charges.

davidlmobley:
ParmEd update is now conda installable, so this passes. Merging.

Issue #189, opened 2016-12-14
Trouble with smarty/smirky "trajectory" files


bannanc:
I had a smirky run on our cluster time out and realized an improvement that should be made on the way the "trajectory" files are output on both smirky and smarty. 

Right now the data for "trajectory" files are saved as a list in the run function instead of just writing to the output file during each step. I don't think this ideal for memory. Also it means we have no trajectory file for simulations that run out of time, which is just wasting computer time. 

I'm working on fixing this on smirky right now. 

@camizanette I can fix this on smarty as well. However, I noticed some other issues with the smarty trajectory file that I thought we had fixed in both samplers, but maybe didn't get moved over to the element decomposition? 

1. In smirky, the trajectory file tracks the total score as well as the score by parameter type created, I thought we were going to add this to smarty as well. 

2. In the smarty trajectories, there are `0` for ParNum and ParentParNum. In smirky I am just using the parameter "typename" in both of these columns with quotes around the name. 

I'm happy to make the changes on the two smarty samplers as well. At the very least before we start any more simulations we should change the way the trajectory output is handled...

davidlmobley:
This sounds good/important, @bannanc . Thanks for working on it.

davidlmobley:
@bannanc - this was resolved by #190, yes?

bannanc:
yes!

PR #190, opened 2016-12-14
Fixing how trajectory files are written

bannanc:
Fixing how trajectory files are written out in smirky (and smarty). 
More details in issue #189

### DIFF
--- a/smarty/sampler.py
+++ b/smarty/sampler.py
@@ -87,7 +87,7 @@ class AtomTypeSampler(object):
         self.typetag = 'atomtype'
 
         # Save bond list to use throughout
-        self.bondset = [("-","simply"), ("=", "doubly"), ("#","triply"), (":", "aromatic"), ("~","any")]
+        self.bondset = [("-","singly"), ("=", "doubly"), ("#","triply"), (":", "aromatic"), ("~","any")]
 
         # Read atomtypes (initial and base) and decorators.
         self.atomtypes = AtomTyper.read_typelist(initialtypes_filename)
@@ -125,6 +125,7 @@ class AtomTypeSampler(object):
             #store empty list of chlidren for each atomtype
             self.parents[smarts] = []
 
+        self.child_to_parent = self._switch_parent_dict()
         # Store a deep copy of the molecules since they will be annotated
         self.molecules = copy.deepcopy(molecules)
 
@@ -568,6 +569,7 @@ class AtomTypeSampler(object):
             self.atomtypes = proposed_atomtypes
             self.molecules = proposed_molecules
             self.parents = proposed_parents
+            self.child_to_parent = self._switch_parent_dict()
             self.atom_type_matches = proposed_atom_type_matches
             self.total_atom_type_matches = proposed_total_atom_type_matches
             return True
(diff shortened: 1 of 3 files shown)


bannanc:
@camizanette 
There is a test failing on the elemental sampler, which I certainly didn't make any changes to. The error message is below:
```
ERROR: test_atomtyper_elemental (smarty.tests.test_sampler.TestAtomTypeSampler)
  File "/home/travis/miniconda/envs/test/lib/python3.4/site-packages/smarty/tests/test_sampler.py", line 58, in test_atomtyper_elemental
    fracfound = atomtype_sampler.run(2)
  File "/home/travis/miniconda/envs/test/lib/python3.4/site-packages/smarty/sampler_elemental.py", line 906, in run
    roots.remove(child)
AttributeError: 'dict_keys' object has no attribute 'remove'
```

bannanc:
Ok, I found what caused this error. I'm a little surprised this is the first time it has come up. However the remove that is causing the error doesn't always happen so if we're only doing 2 iterations in all the tests it might not come up. (heads up @camizanette )

In python2 when you call dict.keys() it returns a list of the keys in that dictionary. 
Instead in python3 it returns an iterator for the keys. 
[Here's a blog post](http://btmiller.com/2015/04/13/get-list-of-keys-from-dictionary-in-python-2-and-3.html) I found while trying to identify what was causing the error incase anyone wants to read more. 

Most of us in the Mobley group are still running python2.7 and were depending on the tests here to catch changes like this.

bannanc:
I would like to include for the record, that the elemental sampler should have inherited almost all of its methods from the original sampler instead of copying code... I'm almost done making fixes to both now.

smarty/sampler.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
         self.typetag = 'atomtype'
 
         # Save bond list to use throughout
-        self.bondset = [("-","simply"), ("=", "doubly"), ("#","triply"), (":", "aromatic"), ("~","any")]
+        self.bondset = [("-","singly"), ("=", "doubly"), ("#","triply"), (":", "aromatic"), ("~","any")]
```
bannanc, line 90: This typo has bothered me for a while and I alway forgot to fix it.

smarty/sampler.py
```diff
@@ -731,6 +743,21 @@ def get_unfinishedAtomList(self, atom_typecounts, molecule_typecounts, atomtype_
 
         return
 
+    def _switch_parent_dict(self):
```
bannanc, line 746: I wanted to be able to get the parent if I knew the child. This is the best I could come up with without completely restructuring the parent dictionary. 
In smirky the parent dictionary is an embedded dictionary so it stores children and parent for every parameter type. I could update smarty to do that instead if we want consistency

smarty/sampler_elemental.py
```diff
@@ -88,13 +88,13 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
         This is just a proof of concept.  No scoring of molecular properties is performed.
 
         """
-        
-        
+
+
```
bannanc, line 92: I have my vim settings delete extra spaces at the end of a line. 
I'm assuming whoever edited this last did not have those settings in place...

davidlmobley: Hmm, that's odd. Everyone who touched it was told to set that, but perhaps someone missed it.

vtlim (approved):
Looks fine to me. (although a bit harder to evaluate out of context)

PR #194, opened 2016-12-19
Add forcefield utility function to merge two parameterized OpenMM systems

nathanmlim:
Forcefield utilities function merge_system will take in two OpenMM systems, assumedly one parameterized with AMBER and another from SMIRFF, use parmed Structures to combine them and regenerates the OpenMM System.


smarty/forcefield_utils.py
```diff
@@ -364,6 +364,50 @@ def getMolParamIDToAtomIndex( oemol, ff):
...
+    label1 (optional) : str
+        String labeling system1 for output. Default, "SMIRFF system"
+    verbose (optional) : bool
+        Print out info on energies, True/False (default True)
+
+    Returns
+    ----------
+    topology : OpenMM Topology
+    system : OpenMM System
+    positions: ???
```
davidlmobley: Says it returns "positions: ???"; it doesn't actually. Strike? (Presumably unneeded?)

smarty/forcefield_utils.py
```diff
@@ -364,6 +364,50 @@ def getMolParamIDToAtomIndex( oemol, ff):
...
+    from simtk.openmm.app import Topology
+    #Merge parameterized Structure
+    structure = structure0 + structure1
+    topology = structure.topology
+
+    #Generate merged OpenMM system
+    system = structure.createSystem()
+
+    if verbose:
+        print("Generating ParmEd Structures...\n \t{}: {}\n \t{}: {}\n".format(label0, structure0, label1, structure1))
```
davidlmobley, line 425: Out of curiosity, do you have an example of what the output from this looks like that you could post in the PR? It's not clear to me.

davidlmobley (commented):
Looks good to me, @limn1 - nice job on the doc strings, commenting reasonably well, etc. I suggested a couple minor tweaks, but other than that, once this passes tests it can be merged.

smarty/forcefield_utils.py
nathanmlim: I wasn't sure if the function should or shouldn't return positions. Does it need to?

davidlmobley: Checking, hold on.

davidlmobley: Hmm, @limn1 - somewhere we will have to get the corresponding positions, as when setting up an actual simulation, one has to set the positions in the context, IIRC. 

Primarily I think it would be just making a single position array (with appropriate indexing) out of the two you would have at the time of input.

It is not totally clear to me that this should be done within the same function or not. If it is, you would want to rename the function, I think. I'd do it within the same function if it makes the bookkeeping easier/avoids code duplication, but separately if it does not.

smarty/forcefield_utils.py
nathanmlim: Looks like this 
```
Generating ParmEd Structures...
 	AMBER system: <Structure 3015 atoms; 239 residues; 2872 bonds; parametrized>
 	SMIRFF system: <Structure 26 atoms; 1 residues; 25 bonds; parametrized>

Merged ParmEd Structure: <Structure 3041 atoms; 240 residues; 2897 bonds; parametrized>
```

smarty/forcefield_utils.py
```diff
@@ -388,25 +392,38 @@ def merge_system( topology0, topology1, system0, system1, label0="AMBER system",
...
+    positions1 = np.array(positions1._value)
+    coordinates = np.vstack((positions0,positions1))
+    natoms = len(coordinates)
+    positions = np.zeros([natoms,3], np.float32)
+    for index in range(natoms):
+        (x,y,z) = coordinates[index]
+        positions[index,0] = x
+        positions[index,1] = y
+        positions[index,2] = z
+    positions = unit.Quantity(positions, unit.angstroms)
```
davidlmobley: Is this perhaps dangerous, @limn1 ? It looks like you assume that the initial position arrays are in angstroms. Should line 417 be `unit.Quantity(positions, positions0._units)` or some such? And a check around 407/408 that both are in angstroms, or put them both in angstroms before using?

smarty/forcefield_utils.py
```diff
@@ -364,6 +364,67 @@ def getMolParamIDToAtomIndex( oemol, ff):
...
+    #Load OpenMM Systems to ParmEd Structures
+    structure0 = parmed.openmm.load_topology( topology0, system0 )
+    structure1 = parmed.openmm.load_topology( topology1, system1 )
+
+    #Merge parameterized Structure
+    structure = structure0 + structure1
+    topology = structure.topology
+
+    #Concatenate positions arrays
+    positions0 = np.array(positions0._value)
```
jchodera: This is unsafe! If you want to convert to pure numpy arrays, use this idiom to safely manage units:
```python
positions_unit = unit.anstroms
positions0_dimensionless = positions0 / positions_unit
positions1_dimensionless = positions1 / positions_unit
# compute `result` somewhere in dimensionless units
positions = unit.Quantity(result_dimensionless, positions_unit)
```

davidlmobley: Thanks, @jchodera . @limn1 - do it this way. :)

davidlmobley:
Excellent, @limn1 , thanks.

PR #198, opened 2017-01-24
Fixing issues with the elemental sampler

bannanc:
This should have been caught in pull request #190 

I'm looking at the tests being run on the elemental sampler to see if I can figure out why this wasn't caught earlier.


davidlmobley:
@bannanc - what's the status on this?

bannanc:
@davidlmobley 
Well the typo that lead to this is fixed, but there seems to be some other possible issues with the elemental sampler. So I think until I can completely debug it we should hold off, I'll write out a detailed issue later today. Assuming no one else is trying to use it right now.

bannanc:
I've actually discovered some significant problems with how the elemental sampler is scoring. Details are in issue #202

bannanc:
@davidlmobley 
when you have time could you do a code review here? 

Please note there is now only one smarty sampler, I added an optional input element. I had a rework a couple of the methods, but now it is definitely consistent sampling rather we look at all atomtypes or 1 specific element. 

I'm going to update the README in the data/molecules/ directory and make sure there are clear examples with both decorator options and how to indicate element examples.

davidlmobley:
@bannanc - thanks! This sounds great. I'll review most likely in the morning.

smarty/sampler.py
```diff
@@ -74,89 +68,141 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
             Temperature for Monte Carlo acceptance/rejection
         verbose : bool, optional, default=False
             If True, verbose output will be printed.
+        decorator_behavior : string either "combinatorial-decorators" or "simple-decorators"
+            simple decorators include bonded atoms as decorators
+        element : iteger >= 0
```
davidlmobley: "iteger" -> "integer".

smarty/sampler.py
```diff
@@ -74,89 +68,141 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
...
-        tmpmolecules = copy.deepcopy(molecules)
-        self.type_molecules(self.basetypes, tmpmolecules)
-        [ basetype_typecounts, molecule_basetype_typecounts] = self.compute_type_statistics( self.basetypes, tmpmolecules )
-
 
         # Compute total atoms
         self.total_atoms = 0.0
         for molecule in self.molecules:
-            for atom in molecule.GetAtoms():
+            for atom in self._GetAtoms(molecule, self.element):
```
davidlmobley, line 194: Ah, clever. Well done.

smarty/sampler.py
```diff
@@ -244,7 +280,9 @@ def best_match_reference_types(self, atomtypes, molecules):
             for reference_atomtype in reference_atomtypes:
                 atoms_in_common[(current_atomtype,reference_atomtype)] = 0
         for (current_typed_molecule, reference_typed_molecule) in zip(molecules, self.reference_typed_molecules):
-            for (current_typed_atom, reference_typed_atom) in zip(current_typed_molecule.GetAtoms(), reference_typed_molecule.GetAtoms()):
+            current_atoms = self._GetAtoms(current_typed_molecule, self.element)
+            reference_atoms = self._GetAtoms(reference_typed_molecule, self.element)
+            for (current_typed_atom, reference_typed_atom) in zip(current_atoms, reference_atoms):
```
davidlmobley, line 285: This is very nice.

smarty/sampler.py
```diff
@@ -317,61 +367,80 @@ def AtomDecorator(self, atom1type, decorator):
...
+        Parameter
+        ---------
+        atom1type : an atomtype tuple (smarts, typename)
+
+        Returns
+        -------
+        True if atomtype has at least 1 alpha substituent otherwise False
         """
+        # TODO: check does this work if you're using replacements
+        # CCB: I don't think it will work!
```
davidlmobley, line 401: "Replacements"? Like macros? Or the recursive smarts?
Does this mean there is a "to do" item here which warrants an issue?

smarty/sampler.py
```diff
@@ -702,6 +774,12 @@ def save_type_statistics(self, typelist, atom_typecounts, molecule_typecounts, a
             output.append("-1,'total','all','None','all',%i,%i,%i,%i" % (ntypes,nmolecules,self.total_atom_type_matches,self.total_atoms))
         return output
 
+    # TODO: CCB - I think this method should be removed, I wrote it initially.
+    # Our goal was to sample only types where all children hadn't been found
+    # I now think this doesn't fit with the idea of sampling where we should
+    # be able to find and remove good things, we shouldn't stop looking at a
+    # branch of the type tree just because each child has a 100% match
+    # currently the resulting list is never used for sampling
```
davidlmobley: I agree with you here, @bannanc .

smarty/tests/test_sampler.py
```diff
@@ -57,14 +56,14 @@ def test_atomtyper_elemental(self):
         """
         Test elemental atomtype sampler for hydrogen
         """
-        atomtype_sampler = smarty.AtomTypeSamplerElemental(self.mols_alkethoh,
+        atomtype_sampler = smarty.AtomTypeSampler(self.mols_alkethoh,
                 self.basetypes, self.basetypes, self.combine_decs,
                 self.replacements, self.mols_alkethoh_ref, 0.1, False,
-                'combinatorial-decorators', '1')
+                'combinatorial-decorators', 1)
```
davidlmobley: Wait, isn't "element" a keyword argument in this function, e.g. the argument should be `element=1` not just `1`??

davidlmobley (approved):
Basically looks good, @bannanc . Once you have your README.md files and examples cleaned up as you mentioned, and you review my (mostly minor) comments and make any changes you deem necessary this is good to merge.

smarty/sampler.py
bannanc: Thanks

smarty/sampler.py
bannanc: @davidlmobley 
I was going to check this before writing an issue. Since we aren't currently using any of the replacement strings in our initial types I didn't think it needed to be addressed in this pull request.

smarty/sampler.py
bannanc: @davidlmobley so should I just remove the method form the script all together?

smarty/tests/test_sampler.py
bannanc: @davidlmobley 
The function is element=0 as a default, I thought if you called a function with the keyword arguments in order you didn't have put the name in front? 
Is there a style preference for this?
The decorator type, temperature, verbose setting, and replacements are also keyword arguments.

smarty/sampler.py
davidlmobley: Sounds good.

smarty/sampler.py
davidlmobley: Yes, IMO.

smarty/tests/test_sampler.py
davidlmobley: Er, I thought the test here said you were trying to test elemental sampling. My mistake, if not. But, you SHOULD test elemental sampling somewhere, right?

bannanc: @davidlmobley this is testing elemental sampling, the element is the last keyword argument so the 1 on the end sets element=1 (instead of 0). 

I'm confused about why I would have to specify element= when calling the function if I'm specifying all of the keyword arguments.

davidlmobley: Ah, I see. OK, stylistically, I find it easier to read what's going on if you say `element=` for an optional argument... But, if that complicates too much you can leave it this way. Your call.

bannanc: OK, that makes more sense, I can change it so its consistent.

davidlmobley: For me it's mostly an issue about argument count. As the argument list gets long and some of them are optimal, I find it hard to track by eyeball which is which unless you use the keywords.

bannanc:
I've made the changes David suggested, assuming I didn't break the test, I'll merge after these tests pass.

bannanc:
Never mind, I took a closer look at my test results from last night, it looks like its still only saving basetypes that are for the element being sampled, it is supposed to save all that match any atom.

bannanc:
There is a molecule in MiniDrugBank that can't be typed by the base-elements, I'm still not quite sure why, I'm going to investigate in the morning. 
If it is an issue with the molecule, I'll update the MiniDrugBank set (AGAIN), but maybe I'm missing something in the basetypes-elemental.smarts file.

davidlmobley:
Interesting, @bannanc ! Keep me posted.

bannanc:
This most recent test failed because the H3 on the "answers" simulation was removed in the only step. I changed the answers test temperature to 0 to prevent this happening in the future. 

I'm looking at MiniDrugBank and it looks like all of the molecules are typed with the elemental basetypes. I'm going to double check the input files, but I thought they were the same as the one I'm testing with.

bannanc:
So, when I run the atomtyper in interactive mode, it can type all molecules in the MiniDrugBank_atyped.oeb file, here is what I called (mirroring how smarty types molecules):
```
from smarty import *
molecules = utils.read_molecules('MiniDrugBank_atyped.oeb')
basetypes = AtomTyper.read_typelist('atomtypes/basetypes-elemental.smarts')
atomtyper = AtomTyper(basetypes, 'atomtype')

# This works without an error
for mol in molecules:
    atomtyper.assignTypes(mol, 0)

# This prints nothing
for mol in molecules:
    for atom in mol.GetAtoms():
        if atom.GetStringData('atomtype') == "":
            print("Found untyped atom")
```

But when I run the smarty calculator with the same in put files I get a `smarty.atomtyper.TypingException`

I'm still investigating why smarty wouldn't work when the interactive does.

bannanc:
@davidlmobley here's an update...

davidlmobley:
@bannanc - are you doing anything different on read, or is there some cleanup being done on the molecules first? Or, interactively, are you somehow loading a different version of smarty? (Perhaps one is being done on the cluster?)

bannanc:
I don't think so, I was interactive on the cluster so I was running the same smarty version. I'm going to remove smarty and re-install in on green planet and run again.

bannanc:
@davidlmobley I found the problem, 

I thought this error was happening with smarty when assigning basetypes, it actually happens earlier than that.

In this round I added a section, to try to limit the bond types to those that are actually used in the molecule set. It appears there might be a chlorine in this molecule that isn't bound to anything. There is apparently a loose Chlorine ion in the molecule, here's a pdf image of the molecule.

[MiniDrugBank_113.pdf](https://github.com/open-forcefield-group/smarty/files/781703/MiniDrugBank_113.pdf)

davidlmobley:
Yes, that would do it. Nice work.

bannanc:
@davidlmobley This is all working now, merging. 

It was tedious and more time consuming than I expected, but I think our scripts and sets of molecules are better for it.

davidlmobley:
Thanks, @bannanc . I also removed the [WIP] from the title. :)

PR #203, opened 2017-02-09
Add smirff99Frosst ffxml to package data 

nathanmlim:
Could we include the [smirff99Frosst](https://github.com/open-forcefield-group/smirff99Frosst/blob/master/smirff99Frosst.ffxml) file in with smarty data package.
This way I can just easily use it like:
`forcefield = ForceField(get_data_filename('forcefield/smirff99Frosst.ffxml'))` 

Since smarty is the parser for this file, it seems like we should include it. It would also save me some headaches in trying to source this file on a remote compute host.


davidlmobley:
This is OK for  me. A better solution long-term is to make the force field itself (which has its own repo) conda installable, but I won't get to that super quickly. So we'll go with this and I'll make issues to remind me to deal with that later.

PR #218, opened 2017-03-06
fix date in smirff99Frosst

bannanc:
As @davidlmobley pointed out we didn't update the date in smirff99Frosst.ffxml with the last pull request. 

I added a comment to the smirffishFrcmod that looked like it was created in August with no note about updates.


davidlmobley:
The py35 test is failing because the latest OpenEye beta is unavailable for this hardware so far; has nothing to do with this PR. Merging and will create a separate issue.

PR #227, opened 2017-03-15
Remove repeated molecules in MiniDrugBank

bannanc:
I removed repeated molecules from the MiniDrugBank set. I was accidentally generating more than one conformation for each molecule so they were being written multiple times to the mol2 file, I've updated the `pickMolecules` ipython notebook and saved the new mol2 files to the smarty/data/molecules directory.


davidlmobley (approved):
Thanks for catching and fixing this. Looks good to me, except that Travis is failing for reasons which are unrelated to this PR.

PR #231, opened 2017-03-16
MiniDrugBank with no generics

bannanc:
This new set of MiniDrugBank molecules covers the same chemical space for smarty (same atomtypes). It no longer has molecules with generic parameters from the current version of smirff99Frosst.


davidlmobley (approved):
Thanks, @bannanc . The changes look good to me.

Issue #222, opened 2017-03-13
How to treat base types in smarty


bannanc:
We've realized one reason smarty cannot get 100% on DrugBank is due to the inability to "OR" element numbers together. This stops us from finding electron withdrawing groups for example. My first instinct is to put $ewg groups into the base types as those are groups of atoms we have decided are worth keeping grouped together. These are already in the substitutions SMARTS file. 

However simply including $ewg in the input file won't be enough. We only keep populated basetypes. We do this by assigning the basetypes to the molecules in the same way the parameters are assigned. The hierarchy means you can't have two base or initial types that match the same atom. 

This means if you had the input list:
```
[#1]      H
[#6]      C
[#8]      O
$ewg1 ewg
```
where `$ewg1` is `[#7!-1,#8!-1,#16!-1,#9,#17,#35,#53]` 
The electron withdrawing group would overwrite oxygen. 
A similar issue came up where it would be worthwhile to have `[*]` as a base type. That will allow things like [#6$(*=*)] as a bonding option. 

I think the easiest fix for this is to check if base types can type an atom without overwriting, I do something similar in smirky and will add a code snippet later today.

davidlmobley:
@bannanc - I'm not totally clear on this phrasing:

> I think the easiest fix for this is to check if base types can type an atom without overwriting,... 

By "type without overwriting", you mean just check if base types can type anything, without checking to see whether they would empty the other base types, right? 

Maybe there should be a warning printed or logged in this case, though, as in some cases this will clearly be a good thing (e.g. as in the case of macros) but in others it would be a mistake.

bannanc:
Sorry, I should have been more clear here. 

"type" was the wrong word. I think we should check if a SMARTS pattern matches any atoms in the set. This is different from assigning a type each atom in the set. 

> Maybe there should be a warning printed or logged in this case, though, as in some cases this will clearly be a good thing (e.g. as in the case of macros) but in others it would be a mistake.

This could be good, thought it will require a more complicated code solution I think. 

In existing methods we have two ways of checking smarts matching: 

1. the atomtyper method `AtomTyper.assignTypes(molecules, element)` this assigns each atom StringData at the typetag matching the current atom type name. 

2. I have a code snippet in smirky that keeps any SMIRKS pattern that matches any set of atoms, it has no memory of if the same set of atoms have matched another pattern before.

bannanc:
@camizanette 

Let me know if you need me to point to specific code snippets

[...]

Issue #17, opened 2016-06-21
Store additional by-iteration information for reuse in plots and subsequent analysis


davidlmobley:
@cbayly13 has been monitoring a variety of information and is thinking about plotting this type of info in various ways. He would at least like to have access to, at each iteration:
- number of parameters (patterns) 
- score

He wants to end up with these in a `pandas` data frame to be able to do things like plot the number of patterns versus score, number of patterns versus iteration, score versus iteration, and so on. This has some overlap with issues raised in #3 and would implement one of the ideas in #15 .

The end goal is for him to end up with this in a `pandas` data frame, but I'm ambivalent at this point as to what the intermediate format would be. Currently he's running `awk` on log files (as output to the screen) which doesn't seem like the best longer-term solution.

bannanc:
@davidlmobley 
I think this has been sufficiently addressed with store typenames and parents in the trajectory files so we could use them to rebuild scores for individual types, the total score, parent trees, etc. 
I'm closing for now.

PR #234, opened 2017-03-17
Changing how SMARTY treats base types

camizanette:
Here, I am solving issue #222.

~The code will have to:~
- [ ] ~Check if there are atoms classified as [*] and [ewg]~
- [ ] ~If so, keep them into self.basetypes without overwriting the previous basetypes~

This pull request does the follow:

- [x] Updates how basetypes are checked so that more generic types can be used when generating new atomtypes
- [x] fixes how basetypes are added to the initial type list such that those not in the initial list are added to the top of the atomtype list in the same order they appear in the input file
- [x] Updates basetype files in /smarty/data/atomtypes to include a generic atom `*` and the two electron withdrawing groups defined in parm99/parm@Frosst.
- [x] Update the README.md file

For now, $ewg1 is not working, there is a warning.

> Warning: Unable to parse SMARTS
> Warning: $ewg1
> Warning: ^
> Warning:  OEPrepareSearch() Substructure object is not initialized yet!

I will try @bannanc's code from SMIRKY now.


bannanc:
@camizanette I'm a little confused by this:
> Check if there are atoms classified as [*] and [ewg]
The point isn't to look specifically for these patterns it is to keep any basetype that matches at least one atom (even if it were over written by following patterns. 

Also, on your warning, I didn't include the substitutions in the code snippet I gave you which would be important, you can look at atomtyper for how to use them, or let me know and I'll add it to the code snippet I gave you.

davidlmobley:
Maybe it will make sense for you two to discuss verbally to improve bandwidth. :)

camizanette:
@bannanc, can you specify all possible new base types?

What did you mean by including the substitutions? I haven't use your code snippet, so my warning is not related to your code.

davidlmobley:
@camizanette - regarding your warning, it looks to me like this is saying that `$ewg` is not getting replaced by the SMARTS pattern it stands for (call this a "substitution"). I would check that the code is using substitutions and, if not, add this in. Specifically, Caitlin wrote:

> I didn't include the substitutions in the code snippet I gave you which would be important, you can look at atomtyper for how to use them, or let me know and I'll add it to the code snippet I gave you.

I think what she's saying is that you have to make sure you're using substitutions to deal with `$ewg` and, if you are unclear how to do that, you should see the atomtyper functionality for how to deal with them, or ask her and she can add them to the code snippet she gave you. 

Also, I'm not totally sure I understand your question:
> can you specify all possible new base types?

Are you saying you want a list of all possible base types that could be allowed? Unless I'm missing something, these are configured in your input files, so there are lots of possibilities. Or are you asking for all possible macros (like `$ewg`) and their substitutions, which could be used as base types? Or something else? (It might help if you said what you think you need this for.)

bannanc:
> What did you mean by including the substitutions? I haven't use your code snippet, so my warning is not related to your code.

I mean that any smarts pattern like $ewg needs to be in your substitution input file and if you're going to do a substructure search you have to use the substitutions to update the SMARTS patterns. otherwise it won't understand $ewg. 

> can you specify all possible new base types?

There are literately an infinite possibility of base types, I'm not sure what you're asking for.

camizanette:
@bannanc It seems I haven't understood the problem. Some basic questions to help me understanding the problem. 

- Why does this current approach doesn't work?

- What do you mean by substitution? I still didn't understand it.

Some broader questions about the problem:

> We've realized one reason smarty cannot get 100% on DrugBank is due to the inability to "OR" element numbers together.

Can you provide the rationale of how you ended up with this hypothesis? Some instances of the problem would help.

> My first instinct is to put $ewg groups into the base types as those are groups of atoms we have decided are worth keeping grouped together.

Any reasoning on how you ended up with this "instinct". Any side effect of doing this solution.

bannanc:
@camizanette I'm going to try to answer these here, but if it is still unclear, feel free to call me (360)-901-1858

**The problem in more detail, I'll cross post this in the issue:**
There are atom types that depend on bonding to atoms OR's together. For example, the actually SMARTS string for H1 is `[#1$(*-[#6]$(*-[#7,#8,#9,#16,#17,#35])]`. That is a hydrogen that is bound to a carbon that is bound to a nitrogen, oxygen, fluorine, sulfur, chlorine, OR bromine. Right now there is no way to OR atom descriptors together in SMARTY. So right now it will usually find something like `[#1$(*-[#6]$(*-[#8])]` which matches H1's with oxygen, but misses the others. 

I noticed a somewhat similar issue with sulfur atomtypes where one of them is `[#16$(*=*)]` While we don't necessarily need to make the exact same SMARTS patterns in smarty, it made me realize that we don't have a way of finding this atom has a double bond to "anything". 

**My proposed solution:**
While we could include another option when making changes to the SMARTS patterns in smarty, ORing things together would be difficult with regex and there are not that many examples of needing an OR'd group of atoms. So the simples solution I came up with would be to put these grouped sets into the input basetypes (since that is where we pull atoms to bind when creating new atomtypes). Because of the hierarchical nature of atomtyping, just putting the OR'd atom option in the basetype file will not work, because we only keep populated basetypes. For example if molecules were typed with the following list:
```
[#6]   C
[#1]   H
[#8]   O
[#7,#8,#9,#16,#17,#35]   ewg
```
Then the ewg SMARTS would match all the oxygen atoms and the `#6` atomtype would be removed. If you include the ewg further up in the list then the other atom types will replace it. Hence My code snippet in issue #222 where we would instead check if the base types match any atoms, instead of using them to "type molecules." The same issue would come up if you included `*    any` in the list 


**Substitutions or replacements**
Substitutions or replacements (we use both words and should probably make this consistent) are a set of SMARTS that can be written in short hand and then use the tag `--substitution`. The following list comes from the file [smarty/smarty/data/atomtypes/replacements.smarts](https://github.com/open-forcefield-group/smarty/blob/master/smarty/data/atomtypes/replacements.smarts):
```
% elements
[#1]    hydrogen
[#6]    carbon
[#7]    nitrogen
[#8]    oxygen
[#9]    fluorine
[#15]   phosphorous
[#16]   sulfur
[#17]   chlorine
[#35]   bromine
[#53]   iodine

% electron withdrawing groups
[#7!-1,#8,#16]              ewg2
[#7!-1,#8!-1,#16!-1,$halogen]     ewg1

% halogens
[$smallhals,$largehals]     halogen
[$fluorine,$chlorine]       smallhals
[$bromine,$iodine]          largehals
```
So this means if you put $ewg1 in your basetypes by using replacements smarty would know you meant `[#7!-1,#8!-1,#16!-1,$halogen]`. Examples for how replacements are used can be seen in the atomtyper code. They must be considered anytime you're doing SMARTS comparisons in SMARTY since we've set it up such that users can put whatever they want in the substitution file and the base or initial types files.

bannanc:
If you're using the `--substitution` to include the replacements.smarts input you shouldn't get this error:

>> Warning: Unable to parse SMARTS
Warning: $ewg1
Warning: ^
Warning: OEPrepareSearch() Substructure object is not initialized yet!

@camizanette obviously if you have other suggestions on how to fix this we would be happy to hear it.

camizanette:
@bannanc Do you want something like `[*$(*-[#6]$(*-[#8])]`?

Will both `$ewg1` and `[*]` be in the base types input file?

Do you want to remove `[*]` and `$ewg` before we call `self.type_molecules` for the first time Since we type all molecules with current typelist to ensure that starting types are sufficient. If we have `[*]` there, it will be always sufficient, but very generic.

Why $ewg1 instead of [#7!-1,#8!-1,#16!-1,$halogen]?

bannanc:
> Do you want something like [*$(*-[#6]$(*-[#8])]?

So there are two ideas here:
1. Should smarty be able to handle it, like if someone put [*] in as their initial type? I think sure why not. 
2. Is this actually something we think would be "good"/useful to generate? No, probably not. Maybe we don't let the '[*]' option be added to the initial types if it is only in the basetypes list? Do you have other ideas about how to avoid it?

> Why $ewg1 instead of [#7!-1,#8!-1,#16!-1,$halogen]?

There's no reason why we couldn't just put `[#7!-1,#8!-1,#16!-1,$halogen]` in the basetypes file (or anywhere else), the point with replacements is just that the functionality is there so we should make sure to support it. 

I'll get to looking at the code next.

smarty/data/atomtypes/basetypes.smarts
```diff
@@ -10,3 +10,5 @@
 [#34]   selenium
 [#35]   bromine
 [#53]   iodine
+[*]     any_atom
+[$ewg1] ewg1
```
bannanc: I would put these at the top of the list, that way if someone were to use this file for the initialtypes option the '*' and 'ewg1' would get over written.

smarty/sampler.py
```diff
@@ -97,6 +98,18 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
...
 
+        used_basetypes = list()
+        self.atomtypes_with_no_matches = set()
+        # Check all SMART strings that are used as a base type
+        for (smarts, atom_type) in self.basetypes:
+            check_flag = self.smarts_matches(smarts)
+            if check_flag == True:
+                used_basetypes.append( ( smarts, atom_type) )
+            else:
+                self.atomtypes_with_no_matches.add( smarts )
```
bannanc: Checking if something `== True` is redundant. You can just call:
```
if self.smarts_matches(smarts):
    used_basetypes.append( ( smarts, atom_type) )
else:
    self.atomtypes_with_no_matches.add( smarts )
```

Or if you think it is more clear to define check_flag then you could just say `if check_flag:` on line 106.

smarty/sampler.py
```diff
@@ -131,54 +144,58 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
...
 
-        # Type molecules again with the updated atomtype list
-        self.type_molecules(self.atomtypes, self.molecules, self.element)
+        ## Type molecules again with the updated atomtype list
+        #self.type_molecules(self.atomtypes, self.molecules, self.element)
 
-        # These are atomtypes where not all children have been matched
-        self.unmatched_atomtypes = copy.deepcopy(self.atomtypes)
+        ## These are atomtypes where not all children have been matched
+        #self.unmatched_atomtypes = copy.deepcopy(self.atomtypes)
```
bannanc: @camizanette Why is the initial types section all commented out? It should still behave the same way as before.

bannanc:
I added a few comments its looking ok so far. 
I assume all the print statements and commenting out are just for testing purposes and will be removed or updated to python3 in the end.

camizanette:
@bannanc I just wanted to check with you if I am on the right direction. I still need to clean up everything and make some adjustments.

Thanks

smarty/sampler.py
camizanette: Testing purpose, I forgot to uncomment.

camizanette:
@bannanc @davidlmobley I haven't noticed a significant improvement with this approach running 5000 iterations. However, it seems SMARTY is getting what Caitlin proposed. Some examples:
```
[#16$(*~[$ewg1])]
[#7$(*=[*])]
[#1$(*~[$ewg1])]
[#6$(*=[$ewg1])$(*~[#8])$(*~[#15])]
[#6$(*=[$ewg1])$(*~[#8])]
```
Could you review it for me? Let me know if I need to do anything else.

bannanc:
@camizanette 
I have it on my list to code review this today, I might not get to it till the afternoon. 

@davidlmobley 
What are the chances of getting the GBSA pull request in soon? Should we try to fix the travis tests in this branch so they'll run properly?

davidlmobley:
@bannanc - not sure how soon GBSA pull request can be in, will revisit it soon to assess. As per our conversation earlier, we could get the testing-related changes into this branch.

davidlmobley:
GBSA pull request is in. Updating branch to get it in here. (Make sure to pull down updates!)

camizanette:
Ok

bannanc:
@camizanette 
I haven't gotten to code reviewing, but can you look into why this is failing? With the GBSA pull request the tests should be running fine.

davidlmobley:
Tests are failing because of issues with the elemental sampler on hydrogen, and on the test beginning with correct answers (where the score is ending up higher than 100%). @camizanette - you'll need to investigate.

bannanc:
@camizanette 
I'm going to hold off on reviewing until the tests pass. 
I'm going to be out of office this week, but I should be able to help if necessary.

camizanette:
@davidlmobley @bannanc

smarty/data/atomtypes/basetypes.smarts
```diff
@@ -1,4 +1,6 @@
 % atom types
+[$ewg1] ewg1
+[*]     any_atom
```
bannanc: There may be other grouped atoms we should be including here. 
I think there are actually two electron withdrawing groups ($ewg1 and $ewg2) in parm@frosst. 
I'll take a look to see what makes the most sense for this part.

bannanc: There may be other grouped atoms we should be including here. 
I think there are actually two electron withdrawing groups ($ewg1 and $ewg2) in parm@frosst. 
I'll take a look to see what makes the most sense for this part. 

Shouldn't these be listed as most general to least general? so the `[*]` option would come first.

smarty/sampler.py
```diff
@@ -129,41 +144,16 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
...
         self.type_molecules(self.atomtypes, self.molecules, self.element)
+
         # Compute atomtype statistics on molecules for current atomtype set
         [atom_typecounts, molecule_typecounts] = self.compute_type_statistics(self.atomtypes, self.molecules, self.element)
         if self.verbose:
             print("MATCHED INITIAL TYPES:")
             self.show_type_statistics(self.atomtypes, atom_typecounts, molecule_typecounts)
 
-        # Track only used atomtypes and add unused to atomtypes with no matches
+        ## Track only used atomtypes and add unused to atomtypes with no matches
```
bannanc: Is there a reason why these comments now have an extra `#`?

smarty/sampler.py
```diff
@@ -129,41 +144,16 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
                 self.atomtypes = [(smarts, typename)] + self.atomtypes
```
bannanc: I've been trying to think through if adding in basetypes in this way makes sense. I'm a little concerned that we're getting things out of order. 

So I think what should happen is that you add any "missing" basetypes to the top of the initial atomtype list in the same order as they appeared in your basetype list. For example if you had basetypes:
```
[*] any
[$ewg1] ewg1
[#6] C
[#8] O
[#1] H
```
and initial types:
```
[#1] hydrogen
[#6] carbon
[#8] oxygen
```
The final list should be:
```
[*] any
[$ewg1] ewg1
[#1] hydrogen
[#6] carbon
[#8] oxygen
```
In the current form I think you would get 
```
[$ewg1] ewg1
[*] any
[#1] hydrogen
[#6] carbon
[#8] oxygen
```

@camizanette I'm still thinking about the best way to avoid this problem, thoughts?

smarty/sampler.py
```diff
@@ -97,6 +97,21 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
         # Save bond list to use throughout
         bondset = [("-","singly"), ("=", "doubly"), ("#","triply"), (":", "aromatic")]
 
+        used_basetypes = list()
+        self.atomtypes_with_no_matches = set()
+        # Check all SMART strings that are used as a base type
+        for (smarts, atom_type) in self.basetypes:
+            check_flag = self.smarts_matches(smarts)
```
bannanc: This is a little knit picky, but I'm not sure I understand why this variable is called `check_flag`? It doesn't really matter what we call it, but it would be better if it was clear from the name what it was defining.

smarty/sampler.py
```diff
@@ -97,6 +97,21 @@ def __init__(self, molecules, basetypes_filename, initialtypes_filename, decorat
...
+            if check_flag:
+                # Keep used base types
+                used_basetypes.append( ( smarts, atom_type) )
+            else:
+                # Remove unused base types
+                self.atomtypes_with_no_matches.add( smarts )
+        self.basetypes = copy.deepcopy(used_basetypes)
+        if verbose:
+            print("USED BASE TYPES: %s" % self.basetypes)
```
bannanc: What does this look like when you print? its inserting a list of tuples into a string should we format this in a way that would be easier to read?

bannanc (commented):
@camizanette 

Your code itself looks great. I had a couple of knit picking comments on naming/style. Then a couple of things I think we need to think about before we finalize here. Let me know if you have any questions.

bannanc:
@davidlmobley 

I added some comments, there are a couple of things I think we need to think about before we finalize here you may want to take a look.

davidlmobley:
@bannanc :

> there are a couple of things I think we need to think about before we finalize here you may want to take a look.

Can you be more specific about what you think I should look at? Do you mean the issues relating to which basetypes are listed first? It seems to me like you would want the wildcard listed first, but my memory probably needs refreshing as it's been quite some time since I did anything with this code.

smarty/sampler.py
camizanette: That is not an important print. I can remove it.

smarty/sampler.py
camizanette: > Shouldn't these be listed as most general to least general? so the [*] option would come first.

That should fix this. Unless you want something to order the basetypes list. And if that is true:

> There are literately an infinite possibility of base types, I'm not sure what you're asking for.

classify them is not trivial.

smarty/sampler.py
camizanette: Thanks for pointing this out!

smarty/sampler.py
davidlmobley: I don't understand your comments, @camizanette .

davidlmobley: Specifically, you asked Caitlin to specify all possible new base types. Since any SMARTS pattern could be a base type, and there are an infinite number of possible SMARTS patterns, there are an infinite number of possible base types. What is it you're trying to accomplish by knowing and/or classifying all possible base types? I think we're not understanding why you want this list or why you want to classify them, so you're going to need to explain.

Also, "That should fix this" is unclear to me -- what's the "that" and the "this" that are referred to?

camizanette: @bannanc wants to have the basetypes in order, the most general to the least general. If WE don't do that by hand on the basetype file, I don't know how we can treat them.
When the basetype filename has the basetypes in order, we don't have the problem of [$ewg1] showing before than [*].

> Also, "That should fix this" is unclear to me -- what's the "that" and the "this" that are referred to?

That: "Shouldn't these be listed as most general to least general? so the [*] option would come first."
This: "So I think what should happen is that you add any "missing" basetypes to the top of the initial atomtype list in the same order as they appeared in your basetype list."

Again, if the basetype file is in order, any "missing" basetypes will show to the top of the initial atomtype list.

**This is how the algorithm is doing:**
```
USED BASE TYPES: [('[*]', 'any_atom'), ('[$ewg1]', 'ewg1'), ('[$ewg2]', 'ewg2'), ('[#1]', 'hydrogen'), ('[#6]', 'carbon'), ('[#8]', 'oxygen')]

Added base (generic) type `[*]`, name c_any_atom, to initial types.
Added base (generic) type `[$ewg1]`, name c_ewg1, to initial types.
Added base (generic) type `[$ewg2]`, name c_ewg2, to initial types.
Added base (generic) type `[#1]`, name c_hydrogen, to initial types.
Added base (generic) type `[#6]`, name c_carbon, to initial types.
Added base (generic) type `[#8]`, name c_oxygen, to initial types.

INDEX        ATOMS  MOLECULES                                                          TYPE NAME                           SMARTS
    1 :          0          0 |                                                       c_any_atom                              [*]
    2 :          0          0 |                                                           c_ewg1                          [$ewg1]
    3 :          0          0 |                                                           c_ewg2                          [$ewg2]
    4 :        464         42 |                                                       c_hydrogen                             [#1]
    5 :        232         42 |                                                         c_carbon                             [#6]
    6 :          0          0 |                                                       c_nitrogen                             [#7]
    7 :        107         42 |                                                         c_oxygen                             [#8]
    8 :          0          0 |                                                       c_fluorine                             [#9]
    9 :          0          0 |                                                    c_phosphorous                            [#15]
   10 :          0          0 |                                                         c_sulfur                            [#16]
   11 :          0          0 |                                                       c_chlorine                            [#17]
   12 :          0          0 |                                                       c_selenium                            [#34]
   13 :          0          0 |                                                        c_bromine                            [#35]
   14 :          0          0 |                                                         c_iodine                            [#53]
TOTAL :        803         42
```

davidlmobley: Yes, @camizanette - I think what Caitlin was proposing was to ask the user to list the base types in order from most general to least general in that file. Otherwise, as you note, there will be no way to SORT them from most general to least general since it's not possible to (a) exhaustively list all SMARTS patterns in order to determine how general each is, or (b) rank SMARTS patterns based on "generality". 

So we should put the SMARTS into this file in order based on generality, and update the docs (doc strings and README.md) to indicate this requirement.

bannanc: @camizanette The issue in this part of the code is independent of the input file. We should not assume that users are using our input files. So I voiced two completely separate concerns:

1. When base types are added to initial types I think they should remain in the same order and be put at the top of the list. 

2. Best practices should have base and initial types that are listed from most to least general, though technically users could do whatever they want. 

I think the best fix to this code would be to replace [lines 140-144](https://github.com/open-forcefield-group/smarty/pull/234/files#diff-8bac8103e4476391b1079b6cebaa477bR140) with:
```
# Add any base types not already there to the initial types
initial_smarts = [ smarts for (smarts, name) in self.atomtypes ]
missing_basetypes = list()
for [smarts, typename] in self.basetypes:
    if smarts not in initial_smarts:
         missing_basetypes.append( (smarts, typename) )
self.atomtypes = missing_basetypes + self.atomtypes
```
This would keep any added basetypes at the top and in the same order they were in in the original input.

bannanc: @davidlmobley 
How to update this code was the comment I thought we might need your input on.

davidlmobley: @camizanette - yes, I agree with Caitlin on this change. These are two separate issues. But the code change she just recommended is needed, otherwise the order of added base types may get jumbled upon addition, and they should be preserved and put at the top of the list. (The initial types would be more specialized, typically, whereas base types could be elemental/generic.)

camizanette: Ok

davidlmobley:
@bannanc @camizanette - is there more to do here or are we good to merge? ( @camizanette - when you think good to merge, make sure your check boxes are checked and the WIP is removed from the title. You also need to update to master.)

bannanc:
@camizanette I think this is good to merge once you pull in the changes in master. 

However, your checklist at the top doesn't really make sense. We're not checking for different basetypes. I edited that comment to strike those out and add what you actually did here.

camizanette:
Maybe should I update the README.md file?

bannanc:
@camizanette yes, that would probably be a good idea! 
I was much more concerned with the code.

smarty/sampler.py
bannanc: @camizanette We should show what base types are kept somewhere though. probably something like

```
if verbose:
    print("USED BASE TYPES:")
    for (smarts, typename) in self.basetypes:
        print("%10s %25s" % (smarts, typename))
```

bannanc:
@camizanette Let us know when you think this isn't WIP any more, I don't think we need a lot of change to the READMEs, but I like the added recommendation you added. 

I suggested one print statement fix, just for visual appeal. 

It would be great if we could get this running ASAP so we might have results for David to share at ACS.

bannanc:
@camizanette is this ready to go, or are there README docs you're still working on?

camizanette:
@bannanc I think it is done. Sorry for not updating the title before.

bannanc:
@davidlmobley I'm merging this

Issue #100, opened 2016-08-06
Constraint specification for SMARTY XML ForceField


jchodera:
At some point, we'll need to expand `ForceField` to allow us to specify which bonds or angles should be constrained.

One way to do this is to extend the `ffxml` format to allow this to be specified via SMIRKS. For example:

``` XML
<Constraints>
  <!-- bonds involving hydrogen -->
  <Constraint smirks="[#1:1]-[*:2]"/> <!-- add constraint between atoms 1 and 2 using equilibrium bond distance -->
  <!-- angles involving hydrogen -->
  <Constraint smirks="[#1:1]-[*:2]-[#1:3]"/> <!-- add constraint between atoms 1 and 3 using auto-calculated distance from equilibrium bond and angles -->
</Constraints>
```

davidlmobley:
This seems reasonable to me. I can discuss with Chris on Monday if you want. I'll tag him here but he's probably out for the weekend. ( @cbayly13 ?)

davidlmobley:
@jchodera - this is where we stand on constraints.

jchodera:
OK! This is super easy to implement if we get the thumbs-up.

We might want to also draft an example for rigid waters. Would that be:
```XML
<Constraints>
  <!-- constrain water O-H bond to equilibrium bond length -->
  <Constraint smirks="[#1X1:1]-[#6X2:2]-[#1X1]"/> 
  <!-- constrain water H...H, calculating equilibrium length from H-O-H equilibrium angle and H-O equilibrium bond lengths -->
  <Constraint smirks="[#1X1:1]-[#6X2:2]-[#1X1:3]"/> 
</Constraints>
```

jchodera:
I guess one question is whether we are OK with the number of atoms tagged in the `smirks` string determine whether to use the equilibrium bond length or whether to interpret the `1-2-3` atom tuple as an angle from which to derive an equilibrium constraint length.

davidlmobley:
I'll ask Chris to look at this. This was sitting there from August, so the comment about asking him in the morning is an old one.

cbayly13:
[...]

PR #239, opened 2017-03-22
Initial support for bond constraints

jchodera:
I'm adding part of the spec from https://github.com/open-forcefield-group/smarty/issues/100 so that we can support rigid TIP3P water.


jchodera:
See the constraint spec addition [here](https://github.com/open-forcefield-group/smarty/blob/a76daef05816e191f5d22ea2478016f1b5fb6b5d/The-SMIRFF-force-field-format.md#constraints).

jchodera:
I'm confused by the failures I'm seeing, which say:
```
ERROR: Test creation of System object from propane_methane_butanol_0.2_0.3_0.5.pdb to test Parm@frosst parameters with charge method OECharges_AM1BCCSym
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty/smarty/tests/test_forcefield.py", line 270, in check_system_creation_from_topology
    check_energy_is_finite(system, positions)
  File "/Users/choderaj/github/open-forcefield-group/smarty/smarty/smarty/tests/test_forcefield.py", line 189, in check_energy_is_finite
    context = openmm.Context(system, integrator)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 6481, in __init__
    this = _openmm.new_Context(*args)
Exception: All Forces must agree on whether to use a cutoff

```

davidlmobley:
As per discussion on Slack, tests are failing because of missing CONECT entries in the prepared boxes containing water. These were prepared with out-of-the-box SolvationToolkit (which uses PackMol) as were the other boxes in `smarty`; however, there is apparently a bug in `openmoltools` which results in CONECT entries being omitted whenever one of the molecules involved is water. Troubleshooting.

See https://github.com/choderalab/openmoltools/issues/253

davidlmobley:
I generated some water boxes with proper CONECT entries (via taking FreeSolv inputs for GROMACS to PDB format via OpenMM) and made a couple minor corrections to the testing, so now tests pass locally. I've also added a README.md within the data directories which indicates where the solvated boxes come from. 

I'll deal with the issue with openmoltools separately.

davidlmobley:
Tests pass, @jchodera ! Merging.

Issue #244, opened 2017-03-26
Remove SMIRFF/SMIRNOFF and ChemicalEnvironment remnants from this repo


jchodera:
We've now migrated the SMIRNOFF forcefield to a new repo:

http://github.com/open-forcefield-group/smirnoff

I was able to preserve commit history on all files that made the transition, though renaming `smarty` to `smirnoff` makes it harder to view this history in github.

Once that is stable, we should remove SMIRNOFF from this repo.

davidlmobley:
Excellent, thanks, @jchodera !!

jchodera:
Let's hold on this until I update the new repo according to our slack discussion.

jchodera:
I have completed the bulk of the reorganization in https://github.com/open-forcefield-group/openforcefield. Please go there and comment/contribute PRs!

jchodera:
We have now completed the migration to http://github.com/open-forcefield-group/openforcefield, so should remove SMIRNOFF and ChemicalEnvironment from this repo.

davidlmobley:
Agreed. Just haven't had the time to do anything about it - still catching up after last week.

bannanc:
I have this working on my local machine. I'm working on changing the travis settings so the tests on github will work. 
See pull request #250

davidlmobley:
Thanks, @bannanc .

bannanc:
This was fixed in [openforcefield PR#7](https://github.com/open-forcefield-group/openforcefield/pull/7)

PR #247, opened 2017-04-06
Clean up conversion utilities already moved to openforcefield

davidlmobley:
Partially resolving #244 by removing `utilities` which has been moved to github.com/open-forcefield-group/openforcefield. Much more still to be done, but this removes some key duplication.


davidlmobley:
@bannanc - after opening this, I realized I perhaps went slightly too far, as these two files might actually belong in `smarty`:

> * `filter_molecule_sets/` - code used to filter DrugBank database including some initial subsets that were not used in final smarty/smirky testing
> * `oeb-to-FF-and-tripos-mol2.py` - utility to convert oeb file with parm@Frosst atomtypes to two `*.mol2` files with tripos and parm@frosst atomtypes

Do these belong here, or with the forcefield stuff?

bannanc:
@davidlmobley 

I think the molecule sets probably make the most sense to be here actually, including the oeb to mol2 script. We use them to test the smarty and smirky simulations, but they've also been used when developing the smirnoff force field

davidlmobley:
Thanks, @bannanc . Restored them here. Can you do a quick check when you have a chance then merge? (See also parallel PR https://github.com/open-forcefield-group/openforcefield/pull/10 -- with these two, everything in `utilities` should now be only in one place rather than two.)

bannanc:
> I think the molecule sets probably make the most sense to be here actually, including the oeb to mol2 script. We use them to test the smarty and smirky simulations, but they've also been used when developing the smirnoff force field

@davidlmobley I wrote that thinking I was on the openforcefield repo I think both of those things should be there. I'll see if I can revert both of these by one commit?
I'm not sure how to do that.

Issue #253, opened 2017-05-18
SMIRKY scoring problem caused by repeat molecules


bannanc:
Last week I noticed a problem with scoring smirky simulations on MiniDrugBank. 

For example, in MiniDrugBank there are 363 molecules with 15,564 atoms
If I type with smirnoff99Frosst, there are 3,354 atoms that match n15 (`[*:1]`). 

Here are the scores for the first iteration of a MiniDrugBank simulation with 

```
PROPOSED:
VdW type matches:
0: [*:1]                                                         matches                                                      n15: [#6:1]:     3291 VdW        types matched
3291 / 15259 total VdWs match (21.568 %)
INDEX          VDWS  MOLECULES   TYPE NAME: SMIRKS                                  REF TYPE: SMIRKS                                   FRACTION OF REF TYPED MOLECULES MATCHED
    1 :      15259        355 | 0: [*:1]                                           n15: [#6:1]                                           3291 /    3354 ( 98.122%)
TOTAL :      15259        363 |                                                       3291 /    15259 match (21.568 %)
```

Some key things to notice: 
* 355 Molecules that are matched with `[*:1]` which should have been 363 molecules!
* There are 3,291 VdW matches to `[*:1]` which should have been 3,354 atoms since the generic VdW should match ALL the VdW types in ALL the molecules. 

I did some investigating today, which included running a lot of the methods in smirky in a jupyter notebook to see where the numbers worked and when they didn't. 

The conclusion: **MiniDrugBank has repeated molecules!** and SMIRKY uses smiles strings as keys in a dictionary to track how molecules are typed with the current list. 

Per a discussion on FreeSolv, I know that OECreateIsoSmiStrings don't take into account stereochemistry so I tried replacing all occurrences of those with OEMolToSmiles so see what happened. With that switch, I get a key error which we already "fixed" with [issue#223](https://github.com/open-forcefield-group/smarty/issues/223)

I did a check outside of smirky with OEMolToSmiles and found it matched 360 of 363 molecules which means there repeating molecules in the set, even considering stereochemistry. 

I've been thinking about fixes here, 

* Should we change MiniDrugBank again to remove these repeated molecules?
* SMIRKY should probably do something else to track molecules, maybe use the index for the molecule as a key in the dictionaries instead?

bannanc:
heads up @davidlmobley and @camizanette

davidlmobley:
[...]

Issue #252, opened 2017-05-18
isinstance for file doesn't work in python3


bannanc:
I tried running smirky on my desktop for the first time today since switching to python3.5. 

I get this error when calling smirky:
```
Traceback (most recent call last):
  File "/Users/bannanc/anaconda3/bin/smirky", line 11, in <module>
    load_entry_point('smarty==0.1.6', 'console_scripts', 'smirky')()
  File "/Users/bannanc/anaconda3/lib/python3.5/site-packages/smarty/cli_smirky.py", line 138, in main
    atom_OR_bases = smarty.parse_odds_file(option.atom_OR_bases, verbose)
  File "/Users/bannanc/anaconda3/lib/python3.5/site-packages/smarty/utils.py", line 77, in parse_odds_file
    if isinstance(filename, file):
NameError: name 'file' is not defined
```

The utils function to parse_odds_file is allowed to take a string or a file object. 

If I run the tests they work, they all pass. The smirky sampler tests do use the parse_odds_file function.  However, if I interactively load the smarty utils function and try to load an odds file I get an error. 
I did some googling and it appears that python3 doesn't support the file type you should instead use:

```python
from io import IOBase

if isinstance(file_object, IOBase):
```

I wrote the parse_odds_file function and I thought I copied this isinstance from another method somewhere, but now I can't find it in smarty or openforcefield. 

@davidlmobley do you have any idea why the tests would pass when this function doesn't work in other places?

davidlmobley:
@bannanc : superficially it looks like the code you're looking at is just related to printing of debug information:
```python
if verbose:
        if isinstance(filename, file):
            print("Attempting to parse file '%s'" % filename.name)
        else:
            print("Attempting to parse file '%s'" % filename)
```
and the tests run with `verbose = False` so this is skipped.

bannanc:
Oh, thank you @davidlmobley I should have seen that! 

I think we should change two things 
1. I don't think the extra if verbose section that the top is necessary
2. Even so, we should probably be running at least 1 test with verbose = True just in case there are other places where this sort of thing comes up.

davidlmobley:
SGTM.

bannanc:
in PR #254

Issue #251, opened 2017-05-09
Potential speed up for smirky by removing chemical environment tracking


bannanc:
I will start by saying I'm not sure if this is worth it at this point in the project, but this is definitely worth thinking about in the long run. 

Right now, smirky keeps a list of chemical environments for the parameter it is sampling. It occurred to me yesterday that this is probably fairly memory intensive and unnecessary since it is easy to go from SMIRKS to `ChemicalEnvironments` and back to SMIRKS. Would it speed up the computational time to remove the chemical environments storage? We could instead only store the SMIRKS strings (and typenames) for the parameter list. Then the `create_new_environment` function could take a smirks string and typename, generate an environment, make changes and return a smirks string and new typename. It seems like it might speed things up significantly since right now we are keeping a chemicalEnvironment for every parameter, which is in the 100's for Torsions and each of the environments is a complicated NetworkX graph with a lot of information that could just be tracked as a SMIRKS string. 

@davidlmobley

davidlmobley:
Is there anything quick you can do to test how much this would affect speed? Naively it seems like it would not be worth it at this point for SMIRKY for the paper, but will be helpful down the line. But I guess that depends on how much additional simulation you still need to do and how much the speed difference is.

bannanc:
The fastest test I can foresee doing is making two simple scripts that store lists for many steps where the list is either SMIRKS strings or chemical environments. Then see if there is a substantial time difference. I'm not sure this is worth the effort now, but it seems like we could potentially be doing A LOT more smirky simulations unless we find why smarty isn't getting 100%

davidlmobley:
@bannanc - I think it's worth trying to test the speed difference if you can do it without a huge investment of time.

bannanc:
I forgot to add my notes here. Below I have data for a list where I sometimes added a new object. The first column is with storing smirks strings the second is for storing chemical environments the last is the difference in minutes (that is string column - environment column). 
generic I started the simulation with only a generic torsion initially, short starts with 10 torsions and long starts with 82 torsions for a recent smirky run. 

```
------------------------------  2 Iterations  ------------------------------
               short	1.97e-05	6.54e-05	4.57e-05
                long	1.93e-05	4.58e-04	4.39e-04
             generic	1.34e-05	1.82e-05	4.84e-06


------------------------------  10 Iterations  ------------------------------
               short	7.12e-05	1.16e-04	4.53e-05
                long	8.27e-05	5.40e-04	4.58e-04
[...]

PR #254, opened 2017-05-24
remove repeat molecules from smirky

bannanc:
* Addressing issue #253 to remove repeat molecules in smirky. 
      * Also removing repeat molecules from smarty
* updating the parse_odds_file function per issue #252 including a test on this function that is verbose.
* add jupyter notebook for testing timing as discussed in issue #251 

~@camizanette and @davidlmobley should we do the same in smarty?~ Yes


davidlmobley:
@bannanc - yes, smarty should also not count repeat molecules multiple times (so they should be removed).

Tag me when you're ready for review.

bannanc:
Well I was about to write something about this being ready to go, but I'll need to figure out what I broke in smarty first.

bannanc:
@davidlmobley I think this one is ready now too

davidlmobley (approved):
Looks good to me! Thanks.

Issue #255, opened 2017-06-05
Remove `The-SMIRFF-force-field-format.md` 


bannanc:
[`The-SMIRFF-force-field-format.md`](https://github.com/open-forcefield-group/smarty/blob/master/The-SMIRFF-force-field-format.md) is still in this repository. The updated version for SMIRNOFF is in openforcefield so I think we should remove the duplicate file.

davidlmobley:
Yes, please.

bannanc:
I did this in my last PR, closing.

Issue #36, opened 2016-07-02
No gitignore file


bannanc:
We don't currently have a gitignore file attached to this repository.

davidlmobley:
Closed by #35.

PR #256, opened 2017-09-18
Updating odds for decision making inside SMIRKY

bannanc:
@davidlmobley and @camizanette 

Here is the update to SMIRKY I'm using to run the new simulations. 
I realized I had some changes in my branch from before I left for the summer that I hadn't pushed. None of these were major. 

Here is a list of changes in this pull request: 

* Removing the file `The-SMIRFF-force-field-format.md` (addressed in issue #255 )
* Renaming some instances of "AlkEtOH" to "AlkEthOH" for consistency (discussed in [openforcefield issue #36](https://github.com/open-forcefield-group/openforcefield/issues/36) )
* Changing temperature 0 to cause only increases in score to be accepted instead of the relative_temperature=1 (We discussed this in person last June (or earlier), however I can't find an issue related to it). 
* Odds updated so "add new atom" happens less frequently
    - This also requires the atom being added to has at least one OR or AND decorator
    - It also gaurentees the new atom gets an ORbase
    - These requirements were there before. It should be noted that decorators can be removed which might have lead to some of the odd patterns I saw in the last set of simulations.


bannanc:
I know this isn't passing tests, I have it fixed now I think, I'll push the new changes in a minute

davidlmobley:
I unpinned the conda-build version, which helped. Looks like now I need to update the encrypted OpenEye license as the one which was used (Chodera lab version I believe) has expired.

davidlmobley (approved):
Looks good to me once tests pass. Thanks, @bannanc .

davidlmobley:
Tests pass. Off to the races.

PR #258, opened 2017-11-06
Correct checking for divide by zero risk when parsing "trajectory" file in score_utils

bannanc:
I am making a PR so we can determine why the nightly tests are failing. I made a small change to a line in the README so github would let me make a PR, but it isn't that important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ All tools for implementation of the SMIRNOFF in OpenMM have been moved to the [o
 * `devtools/` - continuous integration and packaging scripts and utilities
 * `oe_license.txt.enc` - encrypted OpenEye license for continuous integration testing
 * `.travis.yml` - travis-ci continuous integration file
-* `utilities/` - some utility functionality relating to the project.
+* `utilities/` - some utility functionality relating to the project, specifically testing the speed of ChemicalEnvironments for sampling in SMIRKY.
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


bannanc:
Here is the error we are getting:

```
ERROR: Test elemental atomtype sampler for hydrogen
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/smarty/tests/test_sampler.py", line 79, in test_atomtyper_elemental
    atomtype_sampler.run(5, traj)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/smarty/sampler.py", line 912, in run
    timeseries = load_trajectory( trajFile )
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/smarty/score_utils.py", line 68, in load_trajectory
    timeseries[iteration][reftype]['fraction'] = timeseries[iteration][reftype][numerator]/den
ZeroDivisionError: float division by zero
```

The `den` here should be the number of reference types. The only time this is zero is if there are no reference types. A situation which should be caught in the code here, is the part of [`score_utils.load_trajectory`](https://github.com/open-forcefield-group/smarty/blob/master/smarty/score_utils.py#L66) that seems to be causing the problem. 
This seems to only be happening in python3, but my guess is that is due in part to the random nature of example simulations.

bannanc:
I'll try to run some tests locally and see if the same thing ever happens there...

bannanc:
I can't reproduce this locally, I tried starting a new conda environment only installing smarty and its dependencies and tests all pass locally.

jchodera:
That's my guess too. I'd just modify [these lines](https://github.com/open-forcefield-group/smarty/blob/master/smarty/score_utils.py#L67-L68) to not divide by zero if `den` is zero.

bannanc:
@jchodera @davidlmobley or @camizanette 
This is probably a better approach to this section of the code, could someone do a review?

davidlmobley (approved):
Thanks, @bannanc .

PR #259, opened 2017-11-22
Updating SMARTY decorators

camizanette:
Just an update of the decorators used on SMARTY.
I added the rings decorators:

```
% Size of smallest ring
r3
r4
r5
r6
% Number of rings
R0
R1
R2
R3
R4
R
```

@bannanc , could you check if we need anything else there?

### DIFF
--- a/smarty/data/atomtypes/new-decorators.smarts
+++ b/smarty/data/atomtypes/new-decorators.smarts
@@ -1,3 +1,15 @@
+% Size of smallest ring
+r3
+r4
+r5
+r6
+% Number of rings
+R0
+R1
+R2
+R3
+R4
+R
 % total connectivity
 X1             connections-1
 X2             connections-2
--- a/smarty/data/atomtypes/new-decorators.smarts
+++ b/smarty/data/atomtypes/new-decorators.smarts
@@ -3,9 +3,8 @@ r3
 r4
 r5
 r6
-% Number of rings
+% Number of ring bonds
 R0
-R1
 R2
 R3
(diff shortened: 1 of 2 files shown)


bannanc:
So 'R' isn't number of rings, it's number of ring bonds. So R1 never exists so you should probably remove it. I mis understood that also when I made the decorator/odds files for smirky.

camizanette:
So we should probably also update the SMIRKY decorators file? I just did a copy and paste of them.

camizanette:
@bannanc could you check this again?

bannanc:
Sorry, I didn't look at this earlier. 
@camizanette which file do you actually use for smarty simulations?
For the record, `R` (has ring bonds) and `!R0` (doesn't have 0 ring bonds) mean the same thing so you're giving that decorate double probability by including both.

bannanc:
So you know @camizanette the odds files in `smarty/data` I consider examples, they don't reflect the files I'm using for simulations where I weight the decorators unevenly.

camizanette:
@bannanc I have been using new-decorators.smarts file to run on SMARTY. I am not including the "!" decorators there.
Even though you are not using the files for your tests, I think they should have the correct decorators names, so I changed them accordingly.

bannanc:
Oh OK, I misread which file you were changing, I agree that they should have the proper label here. I'll approve and merge this.

bannanc (approved):
Looks good to me, hopefully this will improve smarty's success rate.

PR #260, opened 2017-12-01
Change open-forcefield-group to openforcefield

bannanc:
I don't know how necessary this was, but I figured it would be good to have the links here updated to the new group name just in case.


davidlmobley (approved):
Thanks, @bannanc !

Issue #264, opened 2018-01-25
Add code coverage 


bannanc:
We learned how to do code coverage at MolSSI this week, it is really simple to set up assuming you're running tests with pytest. You just add a line to your travis.yml and go to codecov.io to add the repository and get the link for the badge. 

I can look into the ways to do this with nosetests instead or convert our tests over to pytest.

davidlmobley:
I think converting to pytest and getting code coverage going makes a lot of sense, though is more important on`openforcefield` than here.

bannanc:
I think we have an issue on it for openforcefield too. I feel more comfortable using this one for testing the set up since there isn't active development happening.

Issue #262, opened 2018-01-06
Pin openforcefield version


bannanc:
When the new openforcefield version gets updated, it will break smirky, I suggest we just pin the openforcefield version to 0.0.3 in the smarty conda install

davidlmobley:
Agreed, yes.

PR #265, opened 2018-02-08
[WIP] Add coverage to travis tests

bannanc:
I think this is all we have to do to generate coverage reports for tests, if it works here I can do it on openforcefield also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,7 @@ script:
   # Use beta version for partial bond orders
   - pip install --pre -i https://pypi.anaconda.org/openeye/label/beta/simple openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
   # Install openforcefield tools
-  # TODO if changes to openforcefield become less dynamic switch to conda install?
-  - pip install git+https://github.com/openforcefield/openforcefield.git
+  - conda install -c omnia openforcefield=0.0.3
   # Build the recipe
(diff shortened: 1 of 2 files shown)


davidlmobley:
The README.md shows the code coverage badge, but the status is unknown. Is there a way to tell whether it's working before merging?

(Often I've seen code coverage tests/reports as part of a PR, but turning that on -- likely a toggle on GitHub's settings -- perhaps is a separate test. And maybe the badge only reports on coverage of master?)

Your branch is also out-of-date.

bannanc:
Yes, I meant to put a [WIP] on this when I realized there was something up with the tests. The only coverage I've done was with pytest and directly to a master branch so I'm not sure if the report will show up before this is merged into the master.

bannanc:
For the record this is trying to address #262 and #264

bannanc:
I've been revisiting my github PRs and Issues and found this. I think it would be good for our (mine and @camizanette 's) portfolios for this to not be failing. I think the current issues here relate to the openeye license being expired and maybe a version problem with networkx. I'm going to try to get things passing and then maybe pin versions so we know it is reproducible within the environment it was developed in.

bannanc:
Note on that, going to doesn't mean immediately, just means at some point.

davidlmobley:
All for pinning versions; having the tests frequently start failing because of a change in somethign we don't control is frustrating.

